PR #81, opened 2019-03-25
fix(dde-wallpaper): the mode switch widget overlaps with other widgets

zccrs:
https://github.com/linuxdeepin/internal-discussion/issues/1171


dde-wallpaper-chooser/frame.cpp
```diff
@@ -248,18 +246,41 @@ void Frame::reLayoutTools()
 #endif
     }
 }
+#endif
```
BLumia, line 249: 这个 `#endif` 对应的是哪个？

BLumia (approved):

PR #175, opened 2019-05-23
fix: desktop won't start when fileDialogOnly process exist

BLumia:


dde-desktop/main.cpp
```diff
@@ -120,6 +121,11 @@ int main(int argc, char *argv[])
         }
     }
 
+    if (fileDialogOnly && getuid() == 0) {
```
zccrs: getuid() != 0

zccrs (changes requested):

zccrs (dismissed):

zccrs (approved):

PR #176, opened 2019-05-23
fix: sw_64 should not use mips's desktop file

BLumia:
taskID=5338

``` plain
重现步骤
版本号一致：4.8.4.1-1+stable

连续按super+e，神威上只能打开一个计算机窗口；

x86上，可以打开多个

神威上

deepin@deepin-pc:~$ cat /usr/share/applications/dde-file-manager.desktop | grep Exec
Exec=/usr/bin/file-manager.sh %u
Exec=/usr/bin/file-manager.sh --new-window
```


zccrs (changes requested):
不需要更改

PR #182, opened 2019-05-25
fix: "used space" bar doesn't appear if the icons are set to "extra small"

SiamSami:
It's a fix to a bug. First file manager couldn't show the "used space" bar if the icons were set to "extra small". Now the bug has fixed.

It's a fix to this issue, https://github.com/linuxdeepin/developer-center/issues/998
Before,
DeepinScreenshot_dde-desktop_20190525213904
After,
DeepinScreenshot_dde-desktop_20190525213928

### DIFF
--- a/dde-file-manager-lib/views/computerview.cpp
+++ b/dde-file-manager-lib/views/computerview.cpp
@@ -274,8 +274,8 @@ bool ComputerViewItem::eventFilter(QObject *obj, QEvent *event)
 bool ComputerViewItem::event(QEvent *event)
 {
     if (event->type() == QEvent::Resize) {
-        event->ignore();   // If the program tries to resize itself, then it does that incorrectly. So, the resizing event should be ignored.
-        resize(width(), (getIconLabel()->height() + TEXT_LINE_HEIGHT + ICON_MODE_ICON_SPACING + 90));
+        event->~QEvent();
+        resize(width(), getIconLabel()->height()+TEXT_LINE_HEIGHT+ ICON_MODE_ICON_SPACING + 90);
         adjustPosition();
         return true;
     }
--- a/dde-file-manager-lib/views/computerview.cpp
+++ b/dde-file-manager-lib/views/computerview.cpp
@@ -285,8 +285,8 @@ void ComputerViewItem::adjustPosition()
 {
     m_sizeLabel->setFixedWidth(this->width());
     m_sizeLabel->setAlignment(Qt::AlignCenter);
-    m_sizeLabel->move(0, getTextEdit()->y() + getTextEdit()->height());
-    m_progressLine->move((this->width() - m_progressLine->width()) / 2, m_sizeLabel->y() + m_sizeLabel->height() + 3);
+    m_sizeLabel->move(0, getTextEdit()->y() + getTextEdit()->height() - 7);
+    m_progressLine->move((this->width() - m_progressLine->width()) / 2 , m_sizeLabel->y() + m_progressLine->height() + 25);
 }
 
 bool ComputerViewItem::checked() const
--- a/dde-file-manager-lib/views/computerview.cpp
+++ b/dde-file-manager-lib/views/computerview.cpp
@@ -274,7 +274,7 @@ bool ComputerViewItem::eventFilter(QObject *obj, QEvent *event)
 bool ComputerViewItem::event(QEvent *event)
 {
     if (event->type() == QEvent::Resize) {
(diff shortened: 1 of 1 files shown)


SiamSami:
Can anyone review and merge it? @BLumia @felixonmars @hualet @zccrs

BLumia (commented):
Sorry for the late response since we are on weekend & rest..

TBH I don't think this is the proper way to fix the issue, the main problem is the ComputerViewItem have issue on layout/size policy:

PR

For example, the text edit position is wrong when text need multiple line to display, then all the widget position under the text edit widget will be moved down a little bit. Although your fix can make the progressbar doesn't overflow the widget boundary, it's still not the propery way to do imo.

SiamSami:
@BLumia I'm not sure about that problem. We will fix that later. For now, let's just merge this one.

BLumia:
> let's just merge this one.

Sorry I cannot +1 for PR which does not fix the actual problem..

SiamSami:
@BLumia I honestly have no idea about how to fix the actual problem. We don't get that problem when we set the icon size to something else. We only get that when we set it to "extra small".
Can anyone else fix it? @hualet @felixonmars @zccrs

BLumia:
> I honestly have no idea about how to fix the actual problem. We don't get that problem when we set the icon size to something else. We only get that when we set it to "extra small".

Would like to fix it later when we got time to working on this issue. Sorry I'm really busy these day :(

SiamSami:
@BLumia I think the problem is within QLayout itself. I don't think it's a very big issue.

BLumia:
> @BLumia I think the problem is within QLayout itself. I don't think it's a very big issue.

No it's calculating the position and drawing it by itself, so probably there is something wrong with the calculation.

By the way there is another place is very dirty, which calling `resize()` inside the resizeEvent, which should be fixed later:

https://github.com/linuxdeepin/dde-file-manager/blob/80ea04c79c27f5cbfa66f0aeff68697c0bb78b94/dde-file-manager-lib/views/computerview.cpp#L274-L282

SiamSami:
@BLumia Okay just fixed the original problem now. Now can we merge?
DeepinScreenshot_dde-desktop_20190603132807

SiamSami:
@BLumia Can we merge it now? I think I have fixed the main issue now.

dde-file-manager-lib/views/computerview.cpp
```diff
@@ -274,7 +274,7 @@ bool ComputerViewItem::eventFilter(QObject *obj, QEvent *event)
 bool ComputerViewItem::event(QEvent *event)
 {
     if (event->type() == QEvent::Resize) {
-        resize(width(), getIconLabel()->height() + getTextEdit()->height() + ICON_MODE_ICON_SPACING + 45);
+        resize(width(), getIconLabel()->height() + TEXT_LINE_HEIGHT + ICON_MODE_ICON_SPACING + 90);
```
BLumia, line 278: I cannot accept this change since [Calling resize() inside QEvent::Resize can lead to infinite recursion.](https://doc.qt.io/qt-5/qwidget.html#size-prop), so actually this part of code should be rewrite and removed.

BLumia (changes requested):
Thank you for the contribution but I still cannot accept these changes (see inline comment for details). You can keep these changes for personal build / usage but I cannot get it merge now.

SiamSami:
@BLumia Just updated it again. Can we merge it now?

justforlxz (changes requested):
Can not call resize function in resize event. At least you need to determine if the size given by the event is different from the size you adjusted.

BLumia:
I think these whole logic need to be refactor. Calling resize in ResizeEvent should not be used.

SiamSami:
Okay, so what about now? Does ignoring/disabling the "resize event" work?

SiamSami:
@BLumia Does ignoring/disabling the "resize event" work?

BLumia:
> @BLumia Does ignoring/disabling the "resize event" work?

No, calling `resize()` will create a new resize event so it still have that problem.

SiamSami:
@BLumia So does it work if we destroy the main resize event and start a new one?

When the program has been started, it destroys the `resize event `which was called by the program and then starts another resize event with the correct values. When we resize the icons, the previous `resize event` gets destroyed and another `resize event` is called with the correct values.

SiamSami:
Looks like it's out of my skills. Closing...

PR #209, opened 2019-06-15
feat: show more info

SiamSami:
Before:
DeepinScreenshot_dde-desktop_20190615185853
After:
DeepinScreenshot_dde-desktop_20190615185937

### DIFF
--- a/dde-file-manager-lib/dialogs/computerpropertydialog.cpp
+++ b/dde-file-manager-lib/dialogs/computerpropertydialog.cpp
@@ -103,6 +103,7 @@ void ComputerPropertyDialog::initUI()
 
         gridLayout->addWidget(keyLabel, row, 0, Qt::AlignRight | Qt::AlignTop);
         gridLayout->addWidget(valLabel, row, 1, Qt::AlignLeft | Qt::AlignTop);
+        gridLayout->setRowMinimumHeight(row, valLabel->heightForWidth(valLabel->width())-10);
         row++;
     }
 
--- a/dde-file-manager-lib/dialogs/computerpropertydialog.cpp
+++ b/dde-file-manager-lib/dialogs/computerpropertydialog.cpp
@@ -103,7 +103,7 @@ void ComputerPropertyDialog::initUI()
 
         gridLayout->addWidget(keyLabel, row, 0, Qt::AlignRight | Qt::AlignTop);
         gridLayout->addWidget(valLabel, row, 1, Qt::AlignLeft | Qt::AlignTop);
-        gridLayout->setRowMinimumHeight(row, valLabel->heightForWidth(valLabel->width())-10);
+        gridLayout->setRowMinimumHeight(row, valLabel->heightForWidth(gridLayout->columnMinimumWidth(1)));
         row++;
     }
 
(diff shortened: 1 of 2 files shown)


SiamSami:
@BLumia What do you think about this?

SiamSami (commented):
Let's merge

BLumia:
From your screenshot (after):

where is the `x4` in Processor row?

SiamSami:
@BLumia What about now?
DeepinScreenshot_dde-desktop_20190616225606

dde-file-manager-lib/dialogs/computerpropertydialog.cpp
```diff
@@ -100,7 +103,7 @@ void ComputerPropertyDialog::initUI()
 
         gridLayout->addWidget(keyLabel, row, 0, Qt::AlignRight | Qt::AlignTop);
         gridLayout->addWidget(valLabel, row, 1, Qt::AlignLeft | Qt::AlignTop);
-        gridLayout->setRowMinimumHeight(row, valLabel->heightForWidth(gridLayout->columnMinimumWidth(1)));
+        gridLayout->setRowMinimumHeight(row, valLabel->heightForWidth(valLabel->width())-10);
```
BLumia: Is there any reason to use the number `10` or just because this magic number just works?

SiamSami: @BLumia It's just a magic number.

BLumia: Then I cannot accept this since it may still have problem in some other unexpected cases. Sorry.

SiamSami: @BLumia Then what about the problem in the current dde-file-manager?

SiamSami: @BLumia The reason why I used the number 10 is because without substracting by that number we get some extra spaces. We can use some other numbers as well but 10 seems to be the perfect number.
Without substracting by 10:
DeepinScreenshot_dde-desktop_20190618212637
After substracting by 10:
DeepinScreenshot_dde-desktop_20190618213030

BLumia: I still don't think using a magic number is a good idea since this job should be done by using layout correctly. Maybe you can use [GammaRay](https://github.com/KDAB/GammaRay) or something else to debug the layout issue :)

BLumia:
Since this change affacted the product design (adding new feature is good but we still need product manager's agreement), so cc @wangjia12 .

If we can accpet this feature, you can then revert the layout change and do it in another PR if you like.

SiamSami:
@wangjia12 What do you think about this?

SiamSami:
@BLumia @wangjia12 Okay, I reverted the layout. Now can we merge?

BLumia:
> @BLumia @wangjia12 Okay, I reverted the layout. Now can we merge?

https://github.com/linuxdeepin/dde-file-manager/pull/209#issuecomment-503367156

SiamSami:
@wangjia12 So, do you agree on this?

PR #269, opened 2019-12-02
Disable dmr in arm64

hillwoodroc:


BLumia:
Hi and may I know the reason of this patch? The internal CI did package on arm64 and currently we haven't got bug report about the video preview plugin. Thanks.

hillwoodroc:
Deepin Movie can't be built on openSUSE arm64. I guess it doesn't support arm64. So I think we should disable this option on arm64

BLumia:
Maybe we can wait for v20 release and see if the released deepin-movie can build under arm64, if still failed, we'll discuss what should we do next :)

MCredbear:
> Deepin Movie can't be built on openSUSE arm64. I guess it doesn't support arm64. So I think we should disable this option on arm64

nope，I built it successfully on debian arm64.and debian's repository also has dmr's packages

PR #270, opened 2019-12-12
Fix build with Qt 5.14

antonio-rojas:
This hash is already defined in Qt 5.14, don't redefine it.


BLumia:
Hi and thx for the patch.

Since we are using our internal GitLab instance so I had to manually rebase and merged your patch into the codebase, you can see your patch already get merged in https://github.com/linuxdeepin/dde-file-manager/commit/8632b5e3e369160fad348cfb0d0a731347246443 :)

Maybe it's due to the manually rebase, this PR doesn't marked as merged automated. Anyway thanks again for the patch!

PR #300, opened 2020-11-11
fix: remove the duplicated `StartupNotify` key in dde-open.desktop

cheese:
remove the duplicated `StartupNotify` key in dde-open.desktop

Log:
Signed-off-by: Robin Lee <cheeselee@fedoraproject.org>

### DIFF
--- a/dde-file-manager/dde-open.desktop
+++ b/dde-file-manager/dde-open.desktop
@@ -7,6 +7,5 @@ Icon=dde-file-manager
 NoDisplay=true
 StartupNotify=true
 Name=DDE Open Dialog
-StartupNotify=true
 Terminal=false
 Type=Application


justforlxz:
could you rebase uos? thank you.

cheese:
> could you rebase uos? thank you.

Done. But why do you have to wait for the patch author to do the rebase for such a simple commit?

justforlxz:
> > could you rebase uos? thank you.
> 
> Done. But why do you have to wait for the patch author to do the rebase for such a simple commit?

In fact, we have developed a tool to synchronize the pull request of github to the internal gerrit platform. If the submission is not rebase, it is likely to cause synchronization errors. We have not improved this problem.

PR #306, opened 2020-11-22
BUG: laggy background change after switch between workspace

MrMorteZw:
this change will fix laggy switch workspace


max-lvs:
hello， thanks for your PR, but before review the code you can take one issue to repeat this and we can verify it.

dde-desktop/view/backgroundmanager.cpp
```diff
@@ -433,11 +433,13 @@ void BackgroundManager::onResetBackgroundImage()
         qDebug() << sp->name() << "background path" << userPath << "truesize" << trueSize << "devicePixelRatio"
                  << bw->devicePixelRatioF() << pix << "widget" << bw.get();
         pix.setDevicePixelRatio(bw->devicePixelRatioF());
+        //更新壁纸
```
wangchunlin5013, line 436: Update in cycle? This is definitely not a good idea.
In addition，bw->setPixmap(pix) has actually updated the picture.

wangchunlin5013 (changes requested):

itsXuSt:
Long time no update, close.

PR #449, opened 2022-05-20
fix: disable text highlight on computer view

kt286:
**before**

image

**after**

image

### DIFF
--- a/src/dde-file-manager-lib/views/computerviewitemdelegate.cpp
+++ b/src/dde-file-manager-lib/views/computerviewitemdelegate.cpp
@@ -128,7 +128,7 @@ void ComputerViewItemDelegate::paint(QPainter *painter, const QStyleOptionViewIt
         painter->drawPixmap(option.rect.x() + leftmargin, option.rect.y() + topmargin, icon.pixmap(iconsize));
 
         painter->setFont(par->font());
-        painter->setPen(qApp->palette().color((option.state & QStyle::StateFlag::State_Selected) ? QPalette::ColorRole::HighlightedText : QPalette::ColorRole::Text));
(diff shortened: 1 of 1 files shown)


max-lvs:
Thanks for your advice, but this definition from dde-file-mananger product design, your advice will be considered. :+1:

src/dde-file-manager-lib/views/computerviewitemdelegate.cpp
```diff
@@ -129,7 +129,7 @@ void ComputerViewItemDelegate::paint(QPainter *painter, const QStyleOptionViewIt
         painter->drawPixmap(option.rect.x() + leftmargin, option.rect.y() + topmargin, icon.pixmap(iconsize));
 
         painter->setFont(par->font());
-        painter->setPen(qApp->palette().color((option.state & QStyle::StateFlag::State_Selected) ? QPalette::ColorRole::HighlightedText : QPalette::ColorRole::Text));
```
zccrs, line 131: The following talk is from the designer:

````
The text color of selected state should be use QPalette::BrightText.
````

Please try it and upload the screenshot for both dark and light theme.

zccrs (changes requested):

src/dde-file-manager-lib/views/computerviewitemdelegate.cpp
kt286: image

image

zccrs: Yes, it's right.

Johnson-zs (approved):
Thanks

zccrs (approved):

PR #455, opened 2022-06-21
fix: When previewing text documents in the file manager, the file manager crashes after closing the preview window

lxdaxian:
The problem of the order of recycling objects, and the problem of paging display data; adjust the order of object recycling, and the logic of paging display.
    
Log: Optimize the preview text effect on different platforms
    
Bug: https://pms.uniontech.com/bug-view-141049.html
    
Influence: Preview Text Plugin

### DIFF
--- a/src/dde-file-manager-lib/dialogs/filepreviewdialog.cpp
+++ b/src/dde-file-manager-lib/dialogs/filepreviewdialog.cpp
@@ -514,12 +514,14 @@ void FilePreviewDialog::switchToPage(int index)
 
         if (preview) {
             preview->initialize(this, m_statusBar);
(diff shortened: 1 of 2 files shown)


src/dde-file-manager-lib/dialogs/filepreviewdialog.cpp
```diff
@@ -518,8 +518,10 @@ void FilePreviewDialog::switchToPage(int index)
                 break;
             else if (preview->setFileUrl(m_fileList.at(index)))
```
Johnson-zs: 最好统一括号

Johnson-zs (changes requested):
提交重复

Johnson-zs (approved):
Thanks!

PR #459, opened 2022-06-24
feat: add archlinux build

justforlxz:
add archlinux build

Log:


max-lvs (approved):
thanks for your support!

PR #471, opened 2022-07-13
feat: support hide my directories

kt286:
support hide my directories

Log: support hide my directories

### DIFF
--- a/src/dde-file-manager-lib/models/computermodel.cpp
+++ b/src/dde-file-manager-lib/models/computermodel.cpp
@@ -46,6 +46,7 @@
 #include "vault/vaulthelper.h"
 #include "computermodel.h"
 #include "utils.h"
+#include "dfmapplication.h"
 
 #include <QtConcurrent>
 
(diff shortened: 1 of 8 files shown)


src/dde-file-manager-lib/models/computermodel.cpp
```diff
@@ -61,14 +62,17 @@ ComputerModel::ComputerModel(QObject *parent)
     m_nitems = 0;
     m_appEntryWatcher.reset(fileService->createFileWatcher(nullptr, DUrl(APPENTRY_ROOT), this));
 
+    bool disabledMyDirectories = DFMApplication::genericAttribute(DFMApplication::GA_HiddenMyDirectories).toBool();
```
itsXuSt: How should I make them re-display without restarting dde-file-manager after I changed the config?
Your solution is not COMPLETE so I will NOT approve to merge.
Maybe you should watch the change of the attribute and reload the model when the value changed, that will be more acceptable.

itsXuSt:
features may not be approve, you could make an issue first and out POs will take consider of it.
after that your PR will be consider to merge in.
如 Deepin 社区规范，考虑到商业原因，feature 可能不会被接纳，不过可以先提 issue，产品会考虑需求建议，如果产品觉得可行的话到时候才可能考虑你的需求 PR。

PR #529, opened 2022-08-03
fix: UI refresh problem after mount local smb address

shzhuangpub:
If fetched smb network address and local address are pointed to the save share folder, just fetch the network address.

Log: 修复挂载本机smb地址后，在smb目录创建文件，无法显示
Bug: https://pms.uniontech.com/bug-view-145465.html


shzhuangpub:
close

PR #530, opened 2022-08-03
fix: can not drag file when multi user login

Lighto-Ku:
the process can not attached sharedmemory while another process attaching, so it should get url data from the QMimeData in the drag event.

Log: fix drag issue

Bug: https://pms.uniontech.com/bug-view-144135.html


Johnson-zs (approved):

Johnson-zs (changes requested):
rebase

Johnson-zs (dismissed):

PR #581, opened 2022-09-05
fix: avoid hardcode executable path

wineee:
避免硬编码查找路径，提高可移植性。
https://github.com/linuxdeepin/developer-center/issues/3374


wineee:
close，prefer https://github.com/linuxdeepin/dde-file-manager/pull/580

PR #589, opened 2022-09-13
feat: Record log  when copying from a disc

Johnson-zs:
Includes the number of copied files, file name, file size and file category, operation user, host name, and operation time.

Log: Record log  when copying from a disc

Task: https://pms.uniontech.com/task-view-191163.html


itsXuSt (changes requested):
提交中不要添加内网链接

itsXuSt (approved):

PR #625, opened 2022-10-13
feat: [log] add app start and quit log

misonlan:
as title

Log: add log

Task: https://pms.uniontech.com/task-view-201169.html

### DIFF
--- a/src/dde-file-manager/filemanagerapp.cpp
+++ b/src/dde-file-manager/filemanagerapp.cpp
@@ -30,6 +30,7 @@
 #include "interfaces/dfileservices.h"
 #include "shutil/fileutils.h"
 #include "utils/utils.h"
+#include "utils/rlog/rlog.h"
 #include "app/filesignalmanager.h"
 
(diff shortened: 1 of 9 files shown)


src/dde-file-manager/filemanagerapp.cpp
```diff
@@ -159,9 +160,16 @@ void FileManagerApp::initTranslation()
 
 }
 
-void FileManagerApp::lazyRunInitServiceTask()
+void FileManagerApp::lazyRunTask()
```
itsXuSt, line 163: UT 需要同步修改

itsXuSt (changes requested):

src/dde-file-manager/filemanagerapp.cpp
misonlan: 原来没有ut，现在增加了ut

Kakueeen (approved):

itsXuSt (approved):

PR #655, opened 2022-10-28
fix: Fix display problem of drop-down menu in property dialog

shzhuangpub:
Draw the upper edge and bottom edge of drop down menu.

Bug: https://pms.uniontech.com/bug-view-168005.html


GongHeng2017 (approved):
/approve

obslee (approved):

PR #672, opened 2022-11-07
Fix hardcoded path as per linuxdeepin/developer-center#3374

DonaldKellett:
Change absolute path /usr/bin/x-terminal-emulator in src/dde-file-manager-lib/shutil/fileutils.cpp to x-terminal-emulator

Related: linuxdeepin/developer-center#3374


DonaldKellett:
Still got the commit message wrong, whoops

PR #673, opened 2022-11-07
fix: avoid hardcoded exec path

DonaldKellett:
Replace absolute path to executable /usr/bin/x-terminal-emulator in src/dde-file-manager-lib/shutil/fileutils.cpp with just x-terminal-emulator

Related: linuxdeepin/developer-center#3374

Hopefully I got the commit message format correct this time. If so, please review this pull request and approve if there are no outstanding issues, thanks :smile:

### DIFF
--- a/src/dde-file-manager-lib/shutil/fileutils.cpp
+++ b/src/dde-file-manager-lib/shutil/fileutils.cpp
@@ -1220,11 +1220,11 @@ bool FileUtils::isFileManagerSelf(const QString &desktopFile)
 QString FileUtils::defaultTerminalPath()
 {
     const static QString dde_daemon_default_term = QStringLiteral("/usr/lib/deepin-daemon/default-terminal");
-    const static QString debian_x_term_emu = QStringLiteral("/usr/bin/x-terminal-emulator");
+    const static QString debian_x_term_emu = QStandardPaths::findExecutable("x-terminal-emulator");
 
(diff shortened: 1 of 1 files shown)


src/dde-file-manager-lib/shutil/fileutils.cpp
```diff
@@ -1220,11 +1220,11 @@ bool FileUtils::isFileManagerSelf(const QString &desktopFile)
...
     const static QString dde_daemon_default_term = QStringLiteral("/usr/lib/deepin-daemon/default-terminal");
-    const static QString debian_x_term_emu = QStringLiteral("/usr/bin/x-terminal-emulator");
+    const static QString debian_x_term_emu = QStandardPaths::findExecutable("x-terminal-emulator");
 
     if (QFileInfo::exists(dde_daemon_default_term)) {
         return dde_daemon_default_term;
-    } else if (QFileInfo::exists(debian_x_term_emu)) {
+    } else if (!debian_x_term_emu.isEmpty()) {
         return debian_x_term_emu;
     }
```
itsXuSt, line 1231: add `else` block to return the default xterm path makes code more elegant

DonaldKellett: Thanks for the suggestion, implemented :smile:

max-lvs (approved):
thanks，ok for me

PR #682, opened 2022-11-11
Fix preview key problem

liyigang1:
fix: 【N】 The arrow key previews multiple files, one of which is always missed in the first preview

There are five files on the desktop: community forum, test.pdf, uos.png, video.png, zip.zip In the preview status, press the right key on the keypad to preview the file in sequence. When previewing in sequence, there is a test.pdf missing (if you do not press the right key on the keypad, pressing [>] will not cause this problem)

Log: Fix preview key problem
Bug: https://pms.uniontech.com/bug-view-171079.html

### DIFF
--- a/src/dde-file-manager-lib/dialogs/filepreviewdialog.cpp
+++ b/src/dde-file-manager-lib/dialogs/filepreviewdialog.cpp
@@ -356,15 +356,17 @@ bool FilePreviewDialog::eventFilter(QObject *obj, QEvent *event)
         if (!e->isAutoRepeat()) {
             switch (e->key()) {
             case Qt::Key_Left:
-            case Qt::Key_Up:
-                if (!e->isAutoRepeat())
+            case Qt::Key_Up:{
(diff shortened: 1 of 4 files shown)


src/dde-file-manager-lib/dialogs/filepreviewdialog.cpp
```diff
@@ -356,15 +356,29 @@ bool FilePreviewDialog::eventFilter(QObject *obj, QEvent *event)
...
+                    break;
+                }
                 if (!e->isAutoRepeat())
                     previousPage();
                 break;
+            }
             case Qt::Key_Right:
-            case Qt::Key_Down:
+            case Qt::Key_Down: {
+                qint64 cur = QDateTime::currentMSecsSinceEpoch();
```
max-lvs: 同一片代码，是否考虑优化一下

liyigang1: 好的

max-lvs (approved):

PR #685, opened 2022-11-14
fix: smb integrated item display at the incorret position of sidebar

shzhuangpub:
To control that smb mount address like smb://x.x.x.x would be always put at the end of device group.

Bug: https://pms.uniontech.com/bug-view-169935.html


GongHeng2017 (approved):

GongHeng2017 (dismissed):
/approve

misonlan (approved):

PR #691, opened 2022-11-23
fix: Correct a wrong key string for getting smb password type

shzhuangpub:
Correct `passwordsaveFlag` to `passwordSave`

Bug: https://pms.uniontech.com/bug-view-170201.html


GongHeng2017 (approved):
/approve

misonlan (approved):

PR #704, opened 2022-12-07
fix: Some unsupported pictures can be right clicked to set wallpaper

tingwei1123:
Some unsupported pictures can be right clicked to set wallpaper.

Log: This function is only available for supported wallpaper file suffixes.
Bug: https://pms.uniontech.com/bug-view-174357.html


src/dde-file-manager-lib/interfaces/dabstractfileinfo.cpp
```diff
@@ -954,10 +954,11 @@ QVector<MenuAction> DAbstractFileInfo::menuActionList(DAbstractFileInfo::MenuTyp
                            << MenuAction::Separator;
             } else if (isFile()) {
                 if (mimeTypeName().startsWith("image") && isReadable()
-//                        && !mimeTypeName().endsWith("gif")
-                        && !mimeTypeName().endsWith("svg+xml")
-                        && !mimeTypeName().endsWith("raf")
-                        && !mimeTypeName().endsWith("crw")) {
+                        && (mimeTypeName().endsWith("jpeg")
```
Clauszy, line 957: 使用列表讲支持的类型存起来，这里直接查询是否包含
 image/bmp，image/gif等

Clauszy (changes requested):

PR #715, opened 2023-02-21
chore: correct a typo in dde-select-dialog-x11/main.cpp

felixonmars:
Correct a typo in dde-select-dialog-x11/main.cpp


Johnson-zs:
The master branch has been switched to v23

PR #739, opened 2023-03-24
fix: [propertyDialog] trash property dialog show error in desktop

GongHeng2017:
cause:
the dfmplugin-propertydialog lazy load, but the dfmplugin-trashcore not lazy load, so when dfmplugin-trashcore register widget, the event not define. solved:
listen to the plugin-in initializetion completion event.

Log: fix property dialog problem
Bug: https://pms.uniontech.com/bug-view-192569.html


Clauszy (commented):
+1

Johnson-zs (approved):

PR #743, opened 2023-03-27
修复组策略值变更而设置面板不同步的问题

itsXuSt:
modify the 'dfm.samba.permanent' in dconfig editor, and open the setting
dialog, the item 'show offline...' is not sync.
the value is not obtained from dconfig but from the application json. the
value should be detached from application json config.

Log: fix issue about setting value.

Bug: https://pms.uniontech.com/bug-view-191891.html


Johnson-zs (commented):
+1

max-lvs (approved):

PR #748, opened 2023-03-27
修复自动挂载关闭后，手机设备插入不显示的问题

itsXuSt:
when auto mount is disabled.
ignore the auto mount switcher when protocol device added.

Log: fix issue about protocol device auto mount.

Bug: https://pms.uniontech.com/bug-view-192941.html


Johnson-zs (commented):
+1

Johnson-zs (approved):

PR #749, opened 2023-03-27
fix:  [sidebar/drag] drag item from inner group to other group in sideabr

Johnson-zs:
`sourceItem` is error since the correct `parent` value cannot be obtained. save `curDragItem` in `SideBarModel::mimeData`

Log: fix drag bug

Bug: https://pms.uniontech.com/bug-view-192829.html
Bug: https://pms.uniontech.com/bug-view-192807.html


GongHeng2017 (approved):
+1

itsXuSt (approved):

PR #754, opened 2023-03-28
fix: [ui][tag] fix ui and dsetting default and tag paste logic

pppanghu77:
Match several items in the title to  needs

Log: as des
Bug: https://pms.uniontech.com/bug-view-192981.html https://pms.uniontech.com/bug-view-192079.html
 https://pms.uniontech.com/bug-view-192889.html


Johnson-zs (commented):
+1

max-lvs (approved):
+2

PR #755, opened 2023-03-28
fix: [workspace] hide file need refresh

GongHeng2017:
1. the hidden file path error in rootInfo, it is the parent path.
2. the last file can not hide, because error code logic.
3. send virtual path to workspace in vault.
4. the vault watcher set not cache.

Log: sloved problem of workspace
Bug: https://pms.uniontech.com/bug-view-192887.html
         https://pms.uniontech.com/bug-view-193161.html

### DIFF
--- a/src/plugins/filemanager/core/dfmplugin-workspace/models/rootinfo.cpp
+++ b/src/plugins/filemanager/core/dfmplugin-workspace/models/rootinfo.cpp
@@ -11,6 +11,8 @@
 
 #include <dfm-framework/event/event.h>
 
+#include <dfm-io/dfmio_utils.h>
+
 #include <QApplication>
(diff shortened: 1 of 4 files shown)


src/plugins/filemanager/core/dfmplugin-workspace/models/rootinfo.cpp
Lighto-Ku, line 1: startWork中创建新watcher之前应该先检查当前的watcher是否存在，如果存在则需要清理信号链接之后再执行后续操作，建议封一个函数来做这个watcher创建的工作。

Lighto-Ku (commented):
+1

max-lvs (approved):
+2

PR #756, opened 2023-03-29
fix: [wayland/sidebar] drag item to other group in sidebar

Johnson-zs:
handle drop data that ended by `dropMimeData`

Log: fix drag bug in sidebar

Bug: https://pms.uniontech.com/bug-view-192829.html


itsXuSt (approved):
+1

max-lvs (approved):
+2

PR #758, opened 2023-03-29
chore: always compile organizer but don't install.

Clauszy:
1. need to always compile organizer plugins.
2. using INSTALL_DESKTOP_ORGANIZER to control whether instrall it.

Log:


Johnson-zs (approved):
+1

PR #763, opened 2023-03-29
fix:[ui] siderbar eject item cover

pppanghu77:
short the min size

Log: as des.
Bug: https://pms.uniontech.com/bug-view-192489.html


max-lvs (approved):

Johnson-zs (commented):
+1

PR #766, opened 2023-03-30
一些 UI 调整

itsXuSt:
use Normal button rather recommand button.

Log: fix error dialog button type.

Bug: https://pms.uniontech.com/bug-view-193779.html


Lighto-Ku (commented):
+1

max-lvs (approved):

PR #769, opened 2023-03-30
Add asynchronous fileinfo processing

liyigang1:
Add syncfileinfo and asyncfileinfo, adjust the inheritance structure of all fileinfo, and add the signal and subsequent processing after the asynchronous file attribute is successfully obtained


max-lvs (approved):
refine +2

PR #771, opened 2023-03-30
fix: [vault-UI] update vault UI

GongHeng2017:
update vault UI

Log: update vault UI
Bug: https://pms.uniontech.com/bug-view-193743.html https://pms.uniontech.com/bug-view-193737.html


Lighto-Ku (commented):
+1

liyigang1 (approved):

PR #775, opened 2023-03-31
fix: Add proxy and link file asynchronous query signaling

liyigang1:
1。 In asyncfileinfo, modify d to sharedpointer. 2. Add interfaces notifyUrls and setNotifyUrls. After this asynchronous query is completed, send all url signals that require notification 3. When adding a proxy, the proxy sets the notifyurl. The target file for the linked file is set to notifyurl4. Sending a notifyurl signal in the fileinfo helper

Log: Add proxy and link file asynchronous query signaling Bug:

### DIFF
--- a/src/dfm-base/file/local/asyncfileinfo.h
+++ b/src/dfm-base/file/local/asyncfileinfo.h
@@ -16,7 +16,7 @@ namespace dfmbase {
 class AsyncFileInfoPrivate;
 class AsyncFileInfo : public FileInfo
 {
-    AsyncFileInfoPrivate *d = nullptr;
+    QSharedPointer<AsyncFileInfoPrivate> d { nullptr };
 
 public:
     enum class AsyncAttributeID : uint16_t {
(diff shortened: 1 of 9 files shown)


src/dfm-base/file/local/asyncfileinfo.h
```diff
@@ -180,6 +180,8 @@ class AsyncFileInfo : public FileInfo
     virtual QMap<DFMIO::DFileInfo::AttributeExtendID, QVariant> mediaInfoAttributes(DFMIO::DFileInfo::MediaType type, QList<DFMIO::DFileInfo::AttributeExtendID> ids) const override;
     // cache attribute
     virtual void setExtendedAttributes(const FileExtendedInfoType &key, const QVariant &value) override;
+    QList<QUrl> notifyUrls() const;
+    void setNotifyUrls(const QUrl &url);
```
Kakueeen: 接口名称

max-lvs (approved):
+2

PR #776, opened 2023-03-31
fix: [workspace] The scroll bar dispaly state is wrong

GongHeng2017:
The display state of the scroll bar does not need to be set by itself, but is uniformly set by the control center through dtk.

Log: solved UI problem of workspace
Bug: https://pms.uniontech.com/bug-view-193759.html


Lighto-Ku (commented):
+1

Johnson-zs (approved):

PR #806, opened 2023-04-10
fix: [tag] tag display issue

Lighto-Ku:
show tag in positive order in tag editor.

Log: fix tag display issue

Bug: https://pms.uniontech.com/bug-view-168233.html


max-lvs (approved):
+2， tag

GongHeng2017 (approved):
+1

PR #807, opened 2023-04-10
fix: [view] The view does not display correctly

Kakueeen:
as title
log: fix issue
Bug: https://pms.uniontech.com/bug-view-195097.html


GongHeng2017 (commented):
+1

max-lvs (approved):
+2

PR #810, opened 2023-04-10
fix: [taborder] change tab order of smb passwd dialog.

GongHeng2017:
1. as title;
2. set default focus widget to lineedit.

Log: optimize the taborder of smb password dialog.
Bug: https://pms.uniontech.com/bug-view-170441.html


max-lvs (approved):

Kakueeen (approved):
+1

itsXuSt (approved):

PR #813, opened 2023-04-11
feat: add dconfig to disable desktop

Clauszy:
1. add dconfig key dd.disabled in org.deepin.dde.file-manager.
2. if dd.disabled is true, the desktop do not load any plugins.

Log: 
Task: https://pms.uniontech.com/task-view-256381.html

### DIFF
--- a/src/apps/dde-desktop/main.cpp
+++ b/src/apps/dde-desktop/main.cpp
@@ -7,6 +7,10 @@
 #include "config.h"   //cmake
 #include "tools/upgrade/builtininterface.h"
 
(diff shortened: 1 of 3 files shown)


src/apps/dde-desktop/main.cpp
```diff
@@ -7,6 +7,10 @@
 #include "config.h"   //cmake
 #include "tools/upgrade/builtininterface.h"
 
+#include "dfm-base/base/configs/dconfig/dconfigmanager.h"
```
itsXuSt: base 里不都改成 <> 引入了么

Johnson-zs (approved):

PR #816, opened 2023-04-12
fix: [workspace] After copying/cutting/dragging multiple files, only …

GongHeng2017:
…one file is selected

1. as title
2. After copying a single file, it may not be selected.

Log: solved problem of workspace select file.
Bug: https://pms.uniontech.com/bug-view-195697.html https://pms.uniontech.com/bug-view-195387.html


max-lvs (approved):

Kakueeen:
+1

Lighto-Ku (commented):
+1

PR #818, opened 2023-04-12
fix: [workspace/search] search custom filter issue

Lighto-Ku:
1.set current url to search top widget for set filter callback. 
2.call custom filter callback after default filter.

Log: fix filter issue

Bug: https://pms.uniontech.com/bug-view-195509.html


Kakueeen (approved):
+1

max-lvs (approved):
+2

PR #820, opened 2023-04-12
fix: [translate] Add translation

GongHeng2017:
as title

Log: Add translation
Bug: https://pms.uniontech.com/bug-view-195223.html


itsXuSt (approved):

Lighto-Ku (commented):
+1

PR #823, opened 2023-04-13
fix: remove function of install `dde-file-manger-server`

Johnson-zs:
Causes non-graphical session to fail to install

Log:


justforlxz (approved):

GongHeng2017 (commented):
+1

PR #831, opened 2023-04-13
fix: Unable to delete peripheral files

liyigang1:
When obtaining the fileinfo of the iterator, an asynchronous fileinfo was created, resulting in incorrect determination of the number of directories. Modify fileinfo to directly create the same fileinfo.

Log: Unable to delete peripheral files
Bug: https://pms.uniontech.com/bug-view-194699.html


max-lvs (approved):
+2

Lighto-Ku (commented):
+1

PR #832, opened 2023-04-14
fix: Missing files in folder

liyigang1:
When obtaining the fileinfo of the iterator, an asynchronous fileinfo was created, resulting in incorrect determination of the number of directories. Modify fileinfo to directly create the same fileinfo.

Log: Missing files in folder
Bug: https://pms.uniontech.com/bug-view-195709.html


max-lvs (approved):
+2

PR #837, opened 2023-04-14
fix: [io/trash]When there are too many files in the trash, the operation gets stuck

liyigang1:
There are issues with the process of using the drawing interface.
    1. When modifying the received file information, corner markers, and thumbnail signals, only the corresponding areas will be updated.
    2. Modify the completion signal of the extended corner marker, and do not send it if there is none.
    3. The order and sorting displayed first in sortworker are the same, and the interface is not updated.
    4. When receiving changes in the status of the recycle bin, do not refresh all items and use the trashrooturl to update the specified items.
    5. Modify the trash iterator to obtain fileinfo using the one created earlier.

Log: When there are too many files in the trash, the operation gets stuck
Bug: https://pms.uniontech.com/bug-view-189691.html

### DIFF
--- a/src/dfm-base/file/local/asyncfileinfo.cpp
+++ b/src/dfm-base/file/local/asyncfileinfo.cpp
@@ -568,7 +568,8 @@ void AsyncFileInfo::setNotifyUrl(const QUrl &url)
         return;
     }
     QWriteLocker lk(&d->notifyLock);
-    d->notifyUrls.append(url);
+    if (!d->notifyUrls.contains(url))
+        d->notifyUrls.append(url);
 }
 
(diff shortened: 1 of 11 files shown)


itsXuSt (changes requested):
fix: [xxxxx]

Johnson-zs:
1. 不同的修改，分为多个commits提交
2. 为什么把url改成fileinfo，根因？

liyigang1:
> fix: [xxxxx]

已添加

liyigang1:
> 1. 不同的修改，分为多个commits提交
> 2. 为什么把url改成fileinfo，根因？

已修改

Johnson-zs (approved):

PR #853, opened 2023-04-19
fix: The icons of file-select-dialog are fuzzy

Kakueeen:
Should set attribute as Qt::AA_UseHighDpiPixmaps
Log: fix bug
Bug: https://pms.uniontech.com/bug-view-196199.html


GongHeng2017 (commented):
+1

liyigang1 (approved):

Johnson-zs (approved):

PR #881, opened 2023-04-24
fix: [io/leak]Fix memory or resource leaks after re entry

liyigang1:
1. When repairing a large file copy, resources were not released. 2. After repairing a re-entry loop, resources were released

Log: Fix memory or resource leaks after re entry


Johnson-zs (dismissed):

max-lvs (approved):
+2  leaks....

PR #885, opened 2023-04-24
test: [vault] add vault bug UT

GongHeng2017:
1. as title
2. add bug#193745 test, the cause is windoww not stays on top
3. add bug#178305 test, the cause is not register service
4. add bug#177631 test, the cause is not have transparent encryption function


max-lvs (commented):
content is so simple

max-lvs (approved):

GongHeng2017:
> 



> content is so simple
done

itsXuSt (approved):

PR #887, opened 2023-04-25
fix: [UI] fix connectdailog list view

pppanghu77:
Cancel the built-in background and customize it

Log: fix bug
Bug: https://pms.uniontech.com/bug-view-197591.html  https://pms.uniontech.com/bug-view-197605.html


GongHeng2017 (commented):
+1

max-lvs (approved):

PR #889, opened 2023-04-25
test: [ut/devices] ut for devicewatcher.

itsXuSt:
as title.

Log: add ut cases.

### DIFF
--- a/src/plugins/common/dfmplugin-burn/utils/burnjob.cpp
+++ b/src/plugins/common/dfmplugin-burn/utils/burnjob.cpp
@@ -175,7 +175,7 @@ void AbstractBurnJob::finishFunc(bool verify, bool verifyRet)
     }
 
     emit burnFinished(firstJobType, jobSuccess);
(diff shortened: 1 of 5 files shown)


src/plugins/common/dfmplugin-burn/utils/burnjob.cpp
```diff
@@ -175,7 +175,7 @@ void AbstractBurnJob::finishFunc(bool verify, bool verifyRet)
     }
 
     emit burnFinished(firstJobType, jobSuccess);
-    DeviceManager::instance()->ejectBlockDev(curDevId);
+    DeviceManager::instance()->ejectBlockDevAsync(curDevId);
```
Johnson-zs, line 178: 验证下线程中使用它的影响

Johnson-zs (approved):

src/plugins/common/dfmplugin-burn/utils/burnjob.cpp
itsXuSt: 没有问题。

PR #895, opened 2023-04-26
fix: [io/fileinfo cache] file not show 、file contextmenu show wrong state、desktop crash、delete error

liyigang1:
1.The probability of right clicking to copy, cut, rename, or delete a file is grayed out and cannot be clicked
2.The file flashed by and needs to be refreshed to display it
3.The desktop may experience a black screen phenomenon for a few seconds, and the trash may not be cleared
4.Unable to determine the deletion of USB drive files after enabling the long file name function

bug: https://pms.uniontech.com/bug-view-197911.html 
https://pms.uniontech.com/bug-view-197871.html
https://pms.uniontech.com/bug-view-198113.html
https://pms.uniontech.com/bug-view-198113.html

### DIFF
--- a/src/plugins/filemanager/dfmplugin-vault/views/vaultpropertyview/basicwidget.cpp
+++ b/src/plugins/filemanager/dfmplugin-vault/views/vaultpropertyview/basicwidget.cpp
@@ -126,7 +126,7 @@ void BasicWidget::selectFileUrl(const QUrl &url)
 
     fileCount->setVisible(false);
 
-    if (info->isDir()) {
+    if (info->isAttributes(FileInfo::FileIsType::kIsDir)) {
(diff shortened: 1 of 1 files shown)


max-lvs (dismissed):

max-lvs (dismissed):

include/dfm-base/base/schemefactory.h
```diff
@@ -232,7 +232,7 @@ class InfoFactory final : public SchemeFactory<FileInfo>
 
             if (info) {
                 info->refresh();
-                emit InfoCacheController::instance().cacheFileInfo(url, info);
+                emit InfoCacheController::instance().cacheFileInfo(url, info, false);
```
Johnson-zs: 原则上接口不能随便动，今天加个参数，明天又删个参数，影响太大

PR #903, opened 2023-04-27
fix: [infocache]The probability of right clicking to copy, cut, rename, or delete a file is grayed out and cannot be clicked

liyigang1:
The asyncfileinfo used for peripheral interface drawing and the asyncfileinfo used for right-click menu are not the same instance. So asyncfileinfo did not complete the query, and both reading and writing were errors. After the query is completed, it will display correctly. Add and remove cache interfaces, and first remove the information in the current cache interface during iteration.

Log: The probability of right clicking to copy, cut, rename, or delete a file is grayed out and cannot be clicked
Bug: https://pms.uniontech.com/bug-view-197911.html


GongHeng2017 (commented):
+1

Johnson-zs (approved):

PR #914, opened 2023-04-28
fix: [fileinfo/tag]display issue in desktop

Lighto-Ku:
Can't create file info in the process that remove files, because it will cache a invalid file info.

Log: fix display issue
Bug: https://pms.uniontech.com/bug-view-198159.html


Johnson-zs (changes requested):
这是一个workaround的解法，问题肯定会再次出现

PR #955, opened 2023-05-10
fix: [vault] vault fileview register failed

pppanghu77:
cant register when plugin initialize()

Log: as des
Bug: https://pms.uniontech.com/bug-view-199339.html


max-lvs (commented):
先进master 分支，后面我pick过来

PR #956, opened 2023-05-10
fix: [vault] vault fileview register failed

pppanghu77:
cant register when plugin initialize()

Log: as des
Bug: https://pms.uniontech.com/bug-view-199339.html


GongHeng2017 (approved):
+1

max-lvs (approved):

PR #959, opened 2023-05-11
fix: Failed to cut files from the trash

Kakueeen:
When retrieving the original file name of a file from the trash ,the actual information obtained is the file's path

Log: fix bug
Bug: https://pms.uniontech.com/bug-view-199397.html


GongHeng2017 (commented):
+1

PR #971, opened 2023-05-15
fix: [history]During the loading of the mobile MTP directory, accessing other paths may cause the document management to become stuck

liyigang1:
Attempting to access the current MTP again while iterating through files and switching directories while loading the file in MTP resulted in a lag. Modify the interpretation to determine that it is an mtp file and cannot be used when adding it to historystack

Log: During the loading of the mobile MTP directory, accessing other paths may cause the document management to become stuck
Bug: https://pms.uniontech.com/bug-view-199653.html

### DIFF
--- a/src/plugins/filemanager/core/dfmplugin-titlebar/utils/historystack.cpp
+++ b/src/plugins/filemanager/core/dfmplugin-titlebar/utils/historystack.cpp
@@ -211,10 +211,8 @@ bool HistoryStack::needCheckExist(const QUrl &url)
 
 bool HistoryStack::checkPathIsExist(const QUrl &url)
 {
-    if (FileUtils::isGvfsFile(url)) {
-        return !NetworkUtils::instance()->checkFtpOrSmbBusy(url)
-                ? DFMIO::DFile(url).exists()
-                : false;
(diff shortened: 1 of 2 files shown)


Kakueeen (dismissed):

GongHeng2017 (approved):
+1

GongHeng2017 (dismissed):
+1

src/plugins/filemanager/core/dfmplugin-titlebar/utils/historystack.cpp
```diff
@@ -211,7 +211,7 @@ bool HistoryStack::needCheckExist(const QUrl &url)
 
 bool HistoryStack::checkPathIsExist(const QUrl &url)
 {
-    if (FileUtils::isGvfsFile(url)) {
+    if (FileUtils::isGvfsFile(url) && !FileUtils::isMtpFile(url)) {
```
max-lvs: MTP 只是众多协议设备的一种，还有苹果等其他USB设备呢？

liyigang1: > MTP 只是众多协议设备的一种，还有苹果等其他USB设备呢？

已修改

Johnson-zs (approved):

PR #977, opened 2023-05-16
fix: [workspace]too many times that setIndexActive() func be called.

Lighto-Ku:
just need set the index inactive which already be actived when view scrolling.

Log: fix view freezing issue
Bug: https://pms.uniontech.com/bug-view-200133.html

### DIFF
--- a/src/plugins/filemanager/core/dfmplugin-workspace/views/fileview.cpp
+++ b/src/plugins/filemanager/core/dfmplugin-workspace/views/fileview.cpp
@@ -484,11 +484,13 @@ void FileView::updateModelActiveIndex()
 
     const RandeIndex &rande = randeList.first();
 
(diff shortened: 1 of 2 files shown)


Kakueeen (approved):

src/plugins/filemanager/core/dfmplugin-workspace/views/fileview.cpp
```diff
@@ -914,7 +916,7 @@ void FileView::resizeEvent(QResizeEvent *event)
         doItemsLayout();
 
     if (rootIndex().isValid())
-        updateModelActiveIndex();
+        delayUpdateModelActiveIndex();
```
max-lvs, line 919: 减轻、减缓 不属于解决问题；另外，重资源消耗要反观一下当前设计是否合理

PR #983, opened 2023-05-17
fix: [copy] Copy MTP folder to local, the permissions of folder is error

Kakueeen:
1.dfm-io uses `gio` to query the unix::mode field to get file permissions, but this field is not available in MTP file
2.if the permissions is invalid, do not set permissions

Log: fix bug
Bug: https://pms.uniontech.com/bug-view-199607.html

### DIFF
--- a/src/dfm-base/file/local/localfilehandler.cpp
+++ b/src/dfm-base/file/local/localfilehandler.cpp
@@ -433,6 +433,12 @@ bool LocalFileHandler::setPermissions(const QUrl &url, QFileDevice::Permissions 
         return false;
     }
 
+    // if the `permissions` is invalid, do not set permissions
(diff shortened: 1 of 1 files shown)


src/dfm-base/file/local/localfilehandler.cpp
```diff
@@ -433,6 +433,10 @@ bool LocalFileHandler::setPermissions(const QUrl &url, QFileDevice::Permissions
         return false;
     }
 
+    // if the `permissions` is invalid, do not set permissions
```
max-lvs, line 436: 说清楚，备注清楚，包括bug号

liyigang1 (approved):

Clauszy (approved):

PR #985, opened 2023-05-18
在管理员模式下禁用网络挂载

itsXuSt:
as title.
use macro to control whether to enable smb-in-admin.

Log: disable smb mounts in ADMIN mode.

Bug: https://pms.uniontech.com/bug-view-192907.html


Clauszy (commented):
+1

Clauszy (approved):

PR #990, opened 2023-05-18
fix: [io/mimetype]Application icon display error on desktop

liyigang1:
When creating an asynchronous fileinfo, it was determined that the file mimetype had not yet been obtained, so the desktopfileinfo was not created. When the mimetype is invalid, use dmimedatabase to obtain it.

Log: Application icon display error on desktop
Bug: https://pms.uniontech.com/bug-view-199101.html


Kakueeen (dismissed):

src/plugins/common/dfmplugin-utils/global/virtualglobalplugin.cpp
```diff
@@ -13,6 +13,12 @@
 using namespace dfmplugin_utils;
 DFMBASE_USE_NAMESPACE
 
+static QMimeType readFileMimeType(const QUrl &url)
+{
+    static DMimeDatabase base;
```
Clauszy: DMimeDatabase 这东西是不是考虑统一维护。
代码里面到多static DMimeDatabase，冗余的内存占用

Clauszy (changes requested):

Clauszy (approved):

PR #1010, opened 2023-05-22
fix: [icon] fix when desktopfile icon disabled，Correct display

pppanghu77:
isnull cant judge invalid icon

Log: as title
Bug: https://pms.uniontech.com/bug-view-200885.html


src/dfm-base/file/local/desktopfileinfo.cpp
```diff
@@ -172,7 +172,7 @@ QIcon DesktopFileInfo::fileIcon()
     // 临时代码
     d->icon = QIcon::fromTheme(iconName);   // todo(lxs) SyncFileInfo::fileIcon() 统一处理
 
-    if (d->icon.isNull())
+    if (d->icon.name().isEmpty())
```
max-lvs, line 175: 是不是应该先判NULL，再进行逻辑判断，否则要崩溃

max-lvs (approved):

PR #1019, opened 2023-05-24
fix: The memory manager issue of file-manager

Kakueeen:
Use jemalloc to manager memory

Log: fix bug
Bug: https://pms.uniontech.com/bug-view-200853.html


Clauszy (changes requested):
降低性能，不支持使用

PR #1022, opened 2023-05-25
fix: [io/fileinfoquery]Paste the file into the USB drive, but the file is not displayed

liyigang1:
When creating a new file, insert it into the cache and immediately create a new itemfiledata. When creating a fileinfo in the itemfiledata, you may not have obtained the cached fileinfo yet. Therefore, you have created a new one, and the new one will be queried asynchronously. The previous one will be queried, and all subsequent ones will not be queried. All fileinfo attributes are invalid. Modify the same URL to query different fileinfo instances in order

Log: Paste the file into the USB drive, but the file is not displayed
Bug: https://pms.uniontech.com/bug-view-200183.html

### DIFF
--- a/include/dfm-base/interfaces/abstractsortfilter.h
+++ b/include/dfm-base/interfaces/abstractsortfilter.h
@@ -27,11 +27,11 @@ public:
     AbstractSortFilter();
     virtual ~AbstractSortFilter() {}
     // 左边比右边小返回true
-    virtual int lessThan(const FileInfoPointer left, const FileInfoPointer right,
+    virtual int lessThan(const FileInfoPointer &left, const FileInfoPointer &right,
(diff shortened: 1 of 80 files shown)


src/dfm-base/utils/fileinfohelper.h
```diff
@@ -63,6 +64,7 @@ private Q_SLOTS:
     std::atomic_bool stoped { false };
     QThreadPool pool;
     DThreadList<QUrl> queryingList;
+    DThreadList<FileInfoPointer> needquery;
```
max-lvs: 数据结构混用业务，导致业务理解十分困难。可以在异步fileinfo框架上适配新的业务接口完成。

PR #1042, opened 2023-05-29
fix: [io/asyncfileinfo]The file manager occasionally crashes

liyigang1:
The dfileinfo in asynchronous fileinfo is used by multiple threads without protection. Modify and add thread protection for dfileinfo.

Log: The file manager occasionally crashes

### DIFF
--- a/src/dfm-base/file/local/asyncfileinfo.cpp
+++ b/src/dfm-base/file/local/asyncfileinfo.cpp
@@ -274,10 +274,9 @@ QFileDevice::Permissions AsyncFileInfo::permissions() const
 {
     QFileDevice::Permissions ps;
 
-    if (d->dfmFileInfo)
-        ps = static_cast<QFileDevice::Permissions>(
-                static_cast<uint16_t>(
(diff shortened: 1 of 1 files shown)


src/dfm-base/file/local/asyncfileinfo.cpp
```diff
@@ -991,8 +995,12 @@ QString AsyncFileInfoPrivate::sizeFormat() const
 
 QVariant AsyncFileInfoPrivate::attribute(DFileInfo::AttributeID key, bool *ok) const
 {
-    if (dfmFileInfo) {
-        auto value = dfmFileInfo->attribute(key, ok);
+    const_cast<AsyncFileInfoPrivate *>(this)->dfmFileInfoLock.lock();
```
Johnson-zs: 这是什么操作

liyigang1: > 这是什么操作
已修改

Johnson-zs (approved):

PR #1056, opened 2023-05-30
fix: [select-dialog] select show error

GongHeng2017:
1. when the dir is symlink, get the redirect url
2. in gtk, the parent() will be null, so cache it

Log: fix issue
Bug: https://pms.uniontech.com/bug-view-202149.html

### DIFF
--- a/src/plugins/filedialog/filedialogplugin-core/views/filedialog.cpp
+++ b/src/plugins/filedialog/filedialogplugin-core/views/filedialog.cpp
@@ -261,7 +261,17 @@ QFileDialog::ViewMode FileDialog::currentViewMode() const
 
 void FileDialog::setDirectory(const QString &directory)
 {
-    setDirectoryUrl(UrlRoute::fromLocalFile(directory));
+    QUrl url = UrlRoute::fromLocalFile(directory);
+    QString errorString { "" };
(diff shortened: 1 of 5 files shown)


src/plugins/filedialog/filedialogplugin-core/views/filedialog.cpp
```diff
@@ -261,7 +261,13 @@ QFileDialog::ViewMode FileDialog::currentViewMode() const
 
 void FileDialog::setDirectory(const QString &directory)
 {
-    setDirectoryUrl(UrlRoute::fromLocalFile(directory));
+    QUrl url = UrlRoute::fromLocalFile(directory);
+    QString errorString { "" };
+    const FileInfoPointer &info = InfoFactory::create<FileInfo>(url, Global::CreateFileInfoType::kCreateFileInfoSync, &errorString);
```
Johnson-zs, line 266: 指针检查、errMsg没打印日志

Johnson-zs (commented):
不要在一次commit对多个bug进行修改，如果不是同一个根因的话

src/plugins/filedialog/filedialogplugin-core/views/filedialog.cpp
GongHeng2017: done

Johnson-zs (approved):

Lighto-Ku (approved):

PR #1061, opened 2023-05-30
fix: [workspace]select issue

Lighto-Ku:
selectList be clear before call updateSelecteds() func will make the selectList incorrect.

Log: fix select issue
Bug: https://pms.uniontech.com/bug-view-202271.html


Kakueeen (approved):

GongHeng2017 (commented):
+1

liyigang1 (approved):

max-lvs (approved):

PR #1094, opened 2023-06-08
fix: [fileinfo]The icon of the application program changes to a gear shape after being placed on the USB drive

liyigang1:
Error in obtaining suffix for asynchronous file information in the global conversion function transFileInfo of VirtualGlobalPlugin. Modify the path and nameof in asyncfileinfo to determine that the cache is invalid, and use the parent class fileinfo to obtain these strings.

Log: The icon of the application program changes to a gear shape after being placed on the USB drive
Bug: https://pms.uniontech.com/bug-view-203623.html


Clauszy (commented):
+1

Kakueeen (approved):

max-lvs (approved):

PR #1108, opened 2023-06-12
fix: [recent]Realize Thread safety

pppanghu77:
Change the data into a Thread safety container

Log: as des
Bug:


src/plugins/filemanager/core/dfmplugin-recent/utils/recentmanager.cpp
```diff
@@ -214,7 +213,7 @@ void RecentManager::updateRecent()
 void RecentManager::onUpdateRecentFileInfo(const QUrl &url, const QString originPath, qint64 readTime)
```
itsXuSt: 可以顺便改一下。

pppanghu77: 已修改

Johnson-zs (approved):

PR #1109, opened 2023-06-13
fix: [select-dilaog] repeatly show select box, the app crash

GongHeng2017:
problem:
When using vscode or Google browser repeatedly show selection box, the app crash!
solved:
when show the selection box, add the qtimer to delay display the selection box, solve the crash problem.

Log: fix select dialog problem


Clauszy (changes requested):
请找到崩溃原因，使用timer可能只是暂时规避。

PR #1132, opened 2023-06-19
fix: [select-dialog] the recent dir not update namefilter

GongHeng2017:
cause:
the recent new the filview, so can not get the namefilter. solved:
when cd dir, set the namefilter.

Log: fix issue
Bug: https://pms.uniontech.com/bug-view-205485.html


Lighto-Ku (approved):

src/plugins/filedialog/filedialogplugin-core/views/filedialog.cpp
```diff
@@ -396,6 +396,7 @@ void FileDialog::setFileMode(QFileDialog::FileMode mode)
         || d->fileMode == QFileDialog::Directory) {
         // 清理只显示目录时对文件名添加的过滤条件
         dpfSlotChannel->push("dfmplugin_workspace", "slot_Model_SetNameFilter", internalWinId(), QStringList());
+        curNameFilters = QStringList();
```
Kakueeen: clear接口

GongHeng2017: done

Kakueeen (approved):

max-lvs (approved):

PR #1144, opened 2023-06-21
fix: [io]Unable to iterate over files

liyigang1:
When the same url is opened using multiple windows at the same time, there is only one rootinfo corresponding to multiple models, but one model corresponds to one Iterator. All iteration signals of the Iterator are connected to rootinfo, and the correct model is not found. Add a token to find the correct model.

Log: Unable to iterate over files
Bug: https://pms.uniontech.com/bug-view-205505.html


GongHeng2017 (approved):
+1

Lighto-Ku (approved):

max-lvs (approved):

PR #1150, opened 2023-06-25
修复设备会被挂载到 /media/root/ 目录下的问题

itsXuSt:
device is mounted at /media/root/$DISKNAME while root user is not
loggined. the auto mount operation is dealed in server process, and when
an admin-dfm is launched before, the server which service for admin-dfm
is launched too. and device can be mounted by the admin-server when
device plug in.

Log: fix issue about auto mount.

Bug: https://pms.uniontech.com/bug-view-205301.html

### DIFF
--- a/src/dfm-base/utils/universalutils.cpp
+++ b/src/dfm-base/utils/universalutils.cpp
@@ -127,6 +127,19 @@ QString UniversalUtils::userLoginState()
     return state;
 }
 
+quint32 UniversalUtils::currentLoginUser()
+{
+    quint32 uid = UINT32_MAX;
+    QDBusInterface loginManager("org.freedesktop.login1",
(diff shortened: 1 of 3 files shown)


src/dfm-base/utils/universalutils.cpp
```diff
@@ -127,6 +127,19 @@ QString UniversalUtils::userLoginState()
     return state;
 }
 
+quint32 UniversalUtils::currentLoginUser()
+{
+    quint32 uid;
```
Johnson-zs: 初始化为一个invalid value

itsXuSt: done

Johnson-zs (approved):

PR #1181, opened 2023-07-03
fix: [recent/fileinfo]File manager opens, leaves, closes, and occasionally crashes after recent use

liyigang1:
When closing, the system is destructing. Here, the Static variable in the fileinfo of the file information has been modified, and the vector access again is out of bounds. Modify each attribute of fileinfo and pass it in yourself. Add the stop flag in the identifier worker and exit once stopped.

Log: File manager opens, leaves, closes, and occasionally crashes after recent use

### DIFF
--- a/src/plugins/filemanager/core/dfmplugin-recent/files/recentiterateworker.cpp
+++ b/src/plugins/filemanager/core/dfmplugin-recent/files/recentiterateworker.cpp
@@ -49,6 +49,8 @@ void RecentIterateWorker::onRecentFileChanged(const QList<QUrl> &cachedUrls)
 
         const QUrl &url { QUrl(location) };
         auto info = InfoFactory::create<FileInfo>(url, Global::CreateFileInfoType::kCreateFileInfoSync);
+        if (stopped)
+            return;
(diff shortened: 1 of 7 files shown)


src/plugins/filemanager/core/dfmplugin-recent/files/recentiterateworker.cpp
```diff
@@ -67,11 +71,20 @@ void RecentIterateWorker::onRecentFileChanged(const QList<QUrl> &cachedUrls)
     // delete cached recent file when recent file removed
     QList<QUrl> deletedUrls;
     for (const QUrl &url : cachedUrls) {
+        if (stopped)
+            return;
         if (!urlList.contains(url))
             deletedUrls << url;
     }
+    if (stopped)
+        return;
```
pppanghu77: 这里需要这么多返回点吗

pppanghu77 (approved):

src/plugins/filemanager/core/dfmplugin-recent/files/recentiterateworker.cpp
liyigang1: 已处理

Johnson-zs (approved):

PR #1195, opened 2023-07-11
fix: [fileinfo]Copying files occasionally causes file management to crash

liyigang1:
The asynchronous file information query in the file manager is synchronized by the thread opened by the file manager itself, but at this time, a crash occurs when calling dfmio's devicePathFromUrl. Modify the use of Gio's asynchronous query file information interface to query, and cache the current queried file information using asynchronous threads.

Log: Copying files occasionally causes file management to crash
Bug: https://pms.uniontech.com/bug-view-204813.html

### DIFF
--- a/src/dfm-base/utils/fileinfohelper.cpp
+++ b/src/dfm-base/utils/fileinfohelper.cpp
@@ -15,6 +15,7 @@ DFMBASE_USE_NAMESPACE
 FileInfoHelper::FileInfoHelper(QObject *parent)
     : QObject(parent), thread(new QThread), worker(new FileInfoAsycWorker)
 {
+    moveToThread(qApp->thread());
     init();
 }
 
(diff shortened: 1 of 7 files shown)


src/dfm-base/utils/fileinfohelper.cpp
```diff
@@ -116,6 +131,7 @@ FileInfoHelper::~FileInfoHelper()
 FileInfoHelper &FileInfoHelper::instance()
 {
     static FileInfoHelper helper;
+    helper.moveToThread(qApp->thread());
```
max-lvs: FileInfoHelper 构造函数已经有一次 moveToThread(qApp->thread());  这里再来一次 起什么作用？

max-lvs (commented):
较多的cpp-check 问题，需要修改一下

src/dfm-base/utils/fileinfohelper.cpp
liyigang1: 好的，立刻修改

liyigang1:
> 

已处理

max-lvs (approved):

PR #1207, opened 2023-07-17
fix: [io/taskwidget]Error in copying Progress bar information

liyigang1:
Because when copying to a USB drive, the file information is asynchronous and the attributes of the file have not been obtained yet, the displayed data is not accurate. Modify the timing to obtain file attributes, and if it is correct, it will not be obtained.

Log: Error in copying Progress bar information
Bug: https://pms.uniontech.com/bug-view-209653.html

### DIFF
--- a/src/dfm-base/dialogs/taskdialog/taskwidget.cpp
+++ b/src/dfm-base/dialogs/taskdialog/taskwidget.cpp
@@ -7,6 +7,7 @@
 #include <dfm-base/base/schemefactory.h>
 #include <dfm-base/mimetype/dmimedatabase.h>
 #include <dfm-base/utils/fileutils.h>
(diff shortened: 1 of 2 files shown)


src/dfm-base/dialogs/taskdialog/taskwidget.cpp
```diff
@@ -744,6 +735,62 @@ QString TaskWidget::formatTime(qint64 second) const
     return timeString;
 }
 
+bool TaskWidget::setInfoShow(const FileInfoPointer info, const bool isOrg)
```
Johnson-zs: 注意英文语法

liyigang1: 好的

PR #1211, opened 2023-07-17
fix: [io/taskwidget]Error in copying Progress bar information

liyigang1:
Because when copying to a USB drive, the file information is asynchronous and the attributes of the file have not been obtained yet, the displayed data is not accurate. Modify the timing to obtain file attributes, and if it is correct, it will not be obtained.

Log: Error in copying Progress bar information
Bug: https://pms.uniontech.com/bug-view-209653.html

### DIFF
--- a/src/dfm-base/dialogs/taskdialog/taskwidget.h
+++ b/src/dfm-base/dialogs/taskdialog/taskwidget.h
@@ -7,12 +7,14 @@
 
 #include <dfm-base/dfm_base_global.h>
 #include <dfm-base/interfaces/abstractjobhandler.h>
+#include <dfm-base/interfaces/fileinfo.h>
 
 #include <DWaterProgress>
 #include <DIconButton>
 
 #include <QWidget>
 #include <QLabel>
+#include <QTimer>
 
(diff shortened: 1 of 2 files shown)


Kakueeen (approved):

max-lvs (approved):

src/dfm-base/dialogs/taskdialog/taskwidget.h
```diff
@@ -98,6 +102,10 @@ private Q_SLOTS:
     QAtomicInteger<bool> isPauseState { false };   // 是否是暂停状态
     QAtomicInteger<bool> isBtnHidden { false };   // 暂停、停止按钮隐藏
     QAtomicInteger<bool> isShowError { false };   // 处于错误提示状态
+
+    QTimer infoTimer;
```
max-lvs, line 106: QTimer 危险度高不高哈，我这里看不出来 (:100:

liyigang1: 这里全是在主线程处理，没得其他线程，不存在

PR #1220, opened 2023-07-18
chore: Adapt to dfmio and add init interface

liyigang1:
dde-file-manager needs to know that the iterator initialization is complete when iterating over files. Modifications: 1. Add initIterator interface to the AbstractDirIterator class. 2. Add initIterator interface implementations to the LocalDirIterator, VaultFileIterator and SmbShareIterator classes, calling the dfmio class. 3. DEnumerator's initEnumerator interface.3. Modify the TraversalDirThreadManager class to call initialization in the thread execution, and send the iteratorInitFinished signal when it is done

Log: dde-file-manager adapts the initEnumerator interface added by dfmio.

### DIFF
--- a/src/dfm-base/file/local/localdiriterator.h
+++ b/src/dfm-base/file/local/localdiriterator.h
@@ -44,6 +44,7 @@ public:
     void setArguments(const QVariantMap &args) override;
     QList<SortInfoPointer> sortFileInfoList() override;
     bool oneByOne() override;
+    bool initIterator() override;
     DFMIO::DEnumeratorFuture *asyncIterator();
 };
 }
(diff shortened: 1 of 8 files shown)


src/dfm-base/file/local/localdiriterator.h
```diff
@@ -44,6 +44,7 @@ class LocalDirIterator : public AbstractDirIterator
     void setArguments(const QVariantMap &args) override;
     QList<SortInfoPointer> sortFileInfoList() override;
     bool oneByOne() override;
+    bool init() override;
```
max-lvs: init 一般在类设计中作为私有函数，在类构造函数中调用，作为接口对外，是否可以修改一下名字

Johnson-zs (dismissed):

src/dfm-base/file/local/localdiriterator.h
liyigang1: 已处理

Johnson-zs (approved):

PR #1245, opened 2023-07-25
fix: [crash] App crashed when renaming files with special character sets

Kakueeen:
The length of some special characters is 2, causing the cursor to be incorrectly calculated

Log: fix bug
Bug: https://pms.uniontech.com/bug-view-211431.html

### DIFF
--- a/src/plugins/desktop/core/ddplugin-canvas/delegate/itemeditor.cpp
+++ b/src/plugins/desktop/core/ddplugin-canvas/delegate/itemeditor.cpp
@@ -213,27 +213,6 @@ RenameEdit *ItemEditor::createEditor()
     return edit;
 }
 
-bool ItemEditor::processLength(const QString &srcText, int srcPos, QString &dstText, int &dstPos)
-{
-    int editTextCurrLen = textLength(srcText);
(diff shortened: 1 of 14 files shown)


src/plugins/desktop/core/ddplugin-canvas/delegate/itemeditor.cpp
```diff
@@ -218,14 +218,17 @@ bool ItemEditor::processLength(const QString &srcText, int srcPos, QString &dstT
     int editTextCurrLen = textLength(srcText);
     int editTextRangeOutLen = editTextCurrLen - maximumLength();
     if (editTextRangeOutLen > 0 && maximumLength() != INT_MAX) {
-        QVector<uint> list = srcText.toUcs4();
-        QString tmp = srcText;
-        while (textLength(tmp) > maximumLength() && srcPos > 0) {
-            list.removeAt(--srcPos);
-            tmp = QString::fromUcs4(list.data(), list.size());
+        QString leftText = srcText.left(srcPos);
```
Johnson-zs: 4个地方代码一样，可否封装

max-lvs (approved):

PR #1263, opened 2023-07-27
fix: [vault] the translate error

GongHeng2017:
The qt translator can not identify the macro name space, so do not use the macro.

Log: fix issue
Bug: https://pms.uniontech.com/bug-view-211801.html


max-lvs (approved):

max-lvs (commented):
namespace 的修改怎么换来换去的啊

PR #1375, opened 2023-08-22
fix: [io]Baseline degradation for 1000 file off-disk copy performance comparison

liyigang1:
Add a configuration policy to enable peripheral block device synchronization for each write, and perform synchronization once for each off-disk USB flash drive write. Otherwise, all files are synchronized only at the end of this task.

Log: Baseline degradation for 1000 file off-disk copy performance comparison
Bug: https://pms.uniontech.com/bug-view-215793.html

### DIFF
--- a/assets/configs/org.deepin.dde.file-manager.operations.json
+++ b/assets/configs/org.deepin.dde.file-manager.operations.json
@@ -1,0 +1,28 @@
+{
+    "magic":"dsg.config.meta",
+    "version":"1.0",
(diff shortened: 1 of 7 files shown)


Johnson-zs (commented):
你添加了新的组策略文件，但是没安装啊

liyigang1:
> 你添加了新的组策略文件，但是没安装啊

好的，马上修改

Johnson-zs (approved):

PR #1401, opened 2023-08-31
fix: [views] when the computer name is very long,the name field on th…

scx005548:
…e property page is only partially displayed

as title

Log: When the computer name is exceeded, it is followed by an ellipsis.
Bug: https://pms.uniontech.com/bug-view-217341.html

### DIFF
--- a/src/plugins/common/core/dfmplugin-propertydialog/views/computerpropertydialog.cpp
+++ b/src/plugins/common/core/dfmplugin-propertydialog/views/computerpropertydialog.cpp
@@ -129,7 +129,7 @@ void ComputerPropertyDialog::iniThread()
 void ComputerPropertyDialog::computerProcess(QMap<ComputerInfoItem, QString> computerInfo)
 {
     if (computerInfo.contains(ComputerInfoItem::kName))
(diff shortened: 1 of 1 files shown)


src/plugins/common/core/dfmplugin-propertydialog/views/computerpropertydialog.cpp
```diff
@@ -129,7 +129,7 @@ void ComputerPropertyDialog::iniThread()
 void ComputerPropertyDialog::computerProcess(QMap<ComputerInfoItem, QString> computerInfo)
 {
     if (computerInfo.contains(ComputerInfoItem::kName))
-        computerName->setRightValue(computerInfo[ComputerInfoItem::kName]);
+       computerName->setRightValue(computerInfo[ComputerInfoItem::kName],Qt::TextElideMode::ElideRight);
```
Johnson-zs: , 后面没空格，请配置 clang-format 进行代码格式化

scx005548: 已修改

Johnson-zs (approved):

PR #1437, opened 2023-09-14
fix: [server] server crash under pressure test.

itsXuSt:
the test tests the stability when (un)mount huge amount of samba shares
in short time.
and that makes server crash in thread, items has been remove from a map,
and the map is accessed in another thread without any locker protect.

Solution: use the copy of the map in thread to iterate.

Log: fix issue that server crash under pressure tests.

Bug: https://pms.uniontech.com/bug-view-219219.html


Johnson-zs (approved):
避免了数据竞争，但是引入了数据不同步的问题，自行业务上评估影响是否合入

max-lvs (approved):

PR #1489, opened 2023-10-12
fix: [property-dialog] Can not copy full text in location value.

GongHeng2017:
Add the action to copy full text.

Log: fix issue
Bug: https://pms.uniontech.com/bug-view-219513.html


Johnson-zs (commented):
更新下翻译

GongHeng2017:
> 更新下翻译

done

Johnson-zs (approved):

PR #1512, opened 2023-10-23
fix: [workspace]icon size issue

Lighto-Ku:
reset icon size when changing view mode in file view to make icon size correct.

Log: fix display issue
Bug: https://pms.uniontech.com/bug-view-224471.html


max-lvs (commented):
2级 需要补充UT case 覆盖

liyigang1 (approved):

max-lvs (approved):

PR #1544, opened 2023-10-31
feat: [daemon/anything] Run anything service in independent process

wangrong1069:
Using an independent process to run anything service can solve known log conflict problems. When the service runs abnormally, the service can be restored by creating a new process.

Log: handle high CPU usage
Task: https://pms.uniontech.com/task-view-300371.html


re2zero (approved):
looks good.

PR #1561, opened 2023-11-13
fix: Too many open files when creating thumbnails

Kakueeen:
When the file is unstable (the modification time is less than 2 seconds from the current time), the file is re-added to the thumbnail creation queue

Log: fix bug
Bug: https://pms.uniontech.com/bug-view-227647.html


Johnson-zs (approved):

max-lvs (commented):
请先自测，目前还是不通过

PR #1603, opened 2023-12-01
fix: get audio thumbnail from ID3v2 when ffmpeg can't parse

kt286:
git audio thumbnail from ID3v2 when ffmpeg can't parse


Johnson-zs (changes requested):
当前在dfm-base引入taglib不太建议，dfm-base计划需要瘦身，一些共享库期望是像预览插件一样运行时加载

PR #1614, opened 2023-12-08
fix: to support singel process

max-lvs:
fix single process with muilt window

Log: to support singel process with muilt window
Bug: https://pms.uniontech.com/bug-view-230695.html


re2zero (approved):
+1

PR #1620, opened 2023-12-11
fix: Drag a large file to the system disk, but there is no prompt indicating insufficient disk space. The file manager crashes directly

liyigang1:
Check the disk space usage using the Gio interface. The space provided by Gio is sufficient, but in reality, it is not enough. Copying large files also uses mmap, so there will be no errors and crashes directly. Modify the unified device to obtain free space on the device

Log: Drag a large file to the system disk, but there is no prompt indicating insufficient disk space. The file manager crashes directly
Bug: https://pms.uniontech.com/bug-view-233019.html

### DIFF
--- a/src/dfm-base/base/device/deviceproxymanager.cpp
+++ b/src/dfm-base/base/device/deviceproxymanager.cpp
@@ -151,6 +151,26 @@ bool DeviceProxyManager::isMptOfDevice(const QString &filePath, QString &id)
     return !id.isEmpty();
 }
 
+QVariantMap DeviceProxyManager::queryDeviceInfoByPath(const QString &path, bool reload)
+{
+    d->initMounts();
+    QString blkid, rootblkid;
+    for (auto it = d->allMounts.begin(); it != d->allMounts.end(); it++) {
+        if (it.value() == "/") {
+            rootblkid = it.key();
+            continue;
+        }
+        if (path.startsWith(it.value())
+                || (path + "/").startsWith(it.value())) {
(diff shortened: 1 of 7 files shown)


src/dfm-base/base/device/deviceproxymanager.cpp
```diff
@@ -151,6 +151,26 @@ bool DeviceProxyManager::isMptOfDevice(const QString &filePath, QString &id)
     return !id.isEmpty();
 }
 
+QVariantMap DeviceProxyManager::queryBlockInfoByPath(const QString &path, bool reload)
+{
+    d->initMounts();
```
itsXuSt, line 156: 如果能确认是查询块设备，则可以通过 QStorageInfo::device 函数拿到当前路径对应的块设备节点 /dev/xxx ，再通过 DBlockDeviceMonitor 中的 resolveDeviceNode 可以获取到 ID，然后再查该 ID 对应的设备信息。
避免全量遍历。

itsXuSt (changes requested):

src/dfm-base/base/device/deviceproxymanager.cpp
liyigang1: 这里不止查询块设备的剩余还有查询协议设备的，所以才去遍历所有，使用StorageInfo::device会卡顿

max-lvs (dismissed):

max-lvs (approved):

PR #1622, opened 2023-12-12
fix: when Session recovery the first window, the path of the desktop …

max-lvs:
…first-level directory recovery is incorrect

Send the currentUrlChanged signal when opening the directory for the first time

Log: Send the currentUrlChanged signal when opening the directory for the first time
Bug: https://pms.uniontech.com/bug-view-233397.html

### DIFF
--- a/src/dfm-base/widgets/dfmwindow/filemanagerwindow.cpp
+++ b/src/dfm-base/widgets/dfmwindow/filemanagerwindow.cpp
@@ -197,7 +197,6 @@ FileManagerWindow::FileManagerWindow(const QUrl &url, QWidget *parent)
       d(new FileManagerWindowPrivate(url, this))
 {
     initializeUi();
-    initConnect();
 }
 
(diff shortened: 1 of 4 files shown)


src/dfm-base/widgets/dfmwindow/filemanagerwindow.cpp
```diff
@@ -289,6 +289,7 @@ void FileManagerWindow::installWorkSpace(AbstractFrame *w)
         d->workspace->setCurrentUrl(d->currentUrl);
         d->workspace->installEventFilter(this);
         emit this->workspaceInstallFinished();
+        emit currentUrlChanged(d->currentUrl);   //The URL needs to notify the subscribers when the first window opened.
```
Johnson-zs: 这里加信号不合理

Johnson-zs (changes requested):

Lighto-Ku (approved):

Johnson-zs (approved):

PR #1656, opened 2023-12-24
fix: [disc] allow to append files to folders

Worldwidebrine:
`entryInfoList` doesn't list subdirs.

### DIFF
--- a/src/plugins/filemanager/dfmplugin-optical/views/opticalmediawidget.cpp
+++ b/src/plugins/filemanager/dfmplugin-optical/views/opticalmediawidget.cpp
@@ -20,7 +20,7 @@
 #include <dfm-burn/dburn_global.h>
 
 #include <DSysInfo>
-#include <QDir>
+#include <QDirIterator>
 
 using namespace dfmplugin_optical;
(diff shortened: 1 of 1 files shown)


Worldwidebrine:
I have read the CLA Document and I hereby sign the CLA.

Johnson-zs (commented):
Support for burning files to subdirectories of disc is a good idea, but there's a lot of work to be done beyond just this modification, such as drag and drop or copying files to subdirectories. 
Currently, simply silently deleting duplicate files is not a good approach, as it's dangerous to delete user files and will cause GUI blocking when there are many files. I think it's possible to remove the entire code that removes duplicate files, even though it creates a new problem (if a file with the same name exists, the next burn will fail due to a duplicate file), but this can be optimized with subsequent commits

Johnson-zs:
Thank you for your contribution and submitting this PR.However, I noticed that the issue this PR addresses has already been resolved in another PR that I have submitted ([https://github.com/linuxdeepin/dde-file-manager/pull/1707]).

If you have any other issues or contributions, please don't hesitate to let us know. Thank you for your understanding.

Best regards,

PR #1666, opened 2023-12-26
fix: Unable to open pdf file by Okular

Kakueeen:
If the `NoDisplay` field of the desktop file is set to `true`, it is not displayed in the open mode

Log: fix bug
Bug: https://pms.uniontech.com/bug-view-230471.html


Johnson-zs (commented):
还有一个问题是标题open error 内容显示为 no error

Johnson-zs (approved):

PR #1675, opened 2023-12-27
fix: After selecting multiple directories and right clicking to open the vault, the breadcrumbs display the complete path of the vault

liyigang1:
When modifying and opening, check if it is a directory or if it is an internal jump in the document management system

Log: After selecting multiple directories and right clicking to open the vault, the breadcrumbs display the complete path of the vault
Bug: https://pms.uniontech.com/bug-view-214937.html

### DIFF
--- a/src/plugins/common/core/dfmplugin-menu/menuscene/fileoperatormenuscene.cpp
+++ b/src/plugins/common/core/dfmplugin-menu/menuscene/fileoperatormenuscene.cpp
@@ -207,6 +207,21 @@ bool FileOperatorMenuScene::triggered(QAction *action)
                 dpfSignalDispatcher->publish(GlobalEventType::kChangeCurrentUrl, d->windowId, cdUrl);
             }
         } else {
+            // 如果是目录全部是用文管内部事件打开，因为一个目录就是这么处理的，保持一致，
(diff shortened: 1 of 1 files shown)


src/plugins/common/core/dfmplugin-menu/menuscene/fileoperatormenuscene.cpp
```diff
@@ -207,6 +207,19 @@ bool FileOperatorMenuScene::triggered(QAction *action)
                 dpfSignalDispatcher->publish(GlobalEventType::kChangeCurrentUrl, d->windowId, cdUrl);
             }
         } else {
+            for ( auto it = d->selectFiles.begin(); it != d->selectFiles.end();) {
+                auto info = InfoFactory::create<FileInfo>(*it);
+                if (info.isNull() || !info->isAttributes(OptInfoType::kIsDir) ) {
```
max-lvs, line 214: 备注一下，打开文件夹的特殊处理

liyigang1: 已备注

max-lvs (approved):

PR #1706, opened 2024-01-04
fix: When decompressing a compressed file containing 10000 empty files, the dde-file-manager crashes

liyigang1:
When receiving file updates, both the filesortwork thread and the fileinfocache thread call the Gio property acquisition interface simultaneously, causing a crash. Lock the current operation.

Log: When decompressing a compressed file containing 10000 empty files, the dde-file-manager crashes
Bug: https://pms.uniontech.com/bug-view-237837.html


Johnson-zs (commented):
并不能保证dfmFileInfo->attribute的线程安全

PR #1722, opened 2024-01-11
feat: update the dde-filemanager-daemon security rule

max-lvs:
based on https://www.freedesktop.org/wiki/Software/systemd/dbus/ to update the daemon.service setting

Log: update the dde-filemanager-daemon security rule
Task: https://pms.uniontech.com/task-view-316045.html


re2zero (approved):
/approve

PR #1843, opened 2024-03-05
fix: Creating or copying files will not automatically refresh, in vault

liyigang1:
The monitor that started a non-existent directory in vault was not removed, resulting in the failure of subsequent monitors in the same directory

Log: Creating or copying files will not automatically refresh, in vault
Bug: https://pms.uniontech.com/bug-view-245141.html


Johnson-zs (commented):
urlParent 接口被废弃

liyigang1:
> urlParent 接口被废弃

已处理

Johnson-zs (approved):

PR #1958, opened 2024-04-25
fix: [ui] font size large, item text too short.

itsXuSt:
as title.

Log: as title.


src/plugins/filemanager/core/dfmplugin-computer/delegate/computeritemdelegate.cpp
```diff
@@ -244,7 +244,7 @@ void ComputerItemDelegate::paintSmallItem(QPainter *painter, const QStyleOptionV
 
     QFontMetrics fm(fnt);
 
-    const int TextMaxWidth = option.rect.width() - 20;
+    const int TextMaxWidth = option.rect.width();
```
Johnson-zs: kTextMaxWidth

Johnson-zs (approved):

PR #2073, opened 2024-07-01
fix: [new tab] Change translate of new tab

GongHeng2017:
Change translate of new tab

Log: fix issue
Bug: https://pms.uniontech.com/bug-view-261915.html


Johnson-zs (approved):

Johnson-zs (changes requested):
翻译和代码分成两个commits，否则不好cherry pick

PR #2086, opened 2024-07-04
fix: The maximize button does not display properly

yang233000:
Maximizing the document management, then closing and reopening cannot display the maximize button properly

Log: as title

Bug: https://pms.uniontech.com/bug-view-262201.html


src/dfm-base/widgets/dfmwindow/filemanagerwindow.cpp
```diff
@@ -152,6 +152,7 @@ void FileManagerWindowPrivate::loadWindowState()
     if ((FMWindowsIns.windowIdList().isEmpty()) && ((windowState & kNetWmStateMaximizedHorz) != 0 && (windowState & kNetWmStateMaximizedVert) != 0)) {
         // make window to be maximized.
         // the following calling is copyed from QWidget::showMaximized()
+        q->titlebar()->setVisible(true);
```
Johnson-zs, line 155: 不能搞特殊处理

Johnson-zs (changes requested):

PR #2088, opened 2024-07-04
fix: Refreshing remote directory after network outage caused document management to freeze

liyigang1:
Check if remote access is available once during refresh

Log: Refreshing remote directory after network outage caused document management to freeze
Bug: https://pms.uniontech.com/bug-view-262929.html

### DIFF
--- a/src/dfm-base/file/local/localfilehandler.cpp
+++ b/src/dfm-base/file/local/localfilehandler.cpp
@@ -1086,9 +1086,9 @@ bool LocalFileHandlerPrivate::doOpenFiles(const QList<QUrl> &urls, const QString
 
     bool openResult = doOpenFiles(openInfos, openMineTypes);
 
-    bool openMount = doOpenFiles(mountOpenInfos, openMineTypes);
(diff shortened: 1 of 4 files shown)


max-lvs (dismissed):

src/dfm-base/file/local/localfilehandler.cpp
```diff
@@ -1086,7 +1086,7 @@ bool LocalFileHandlerPrivate::doOpenFiles(const QList<QUrl> &urls, const QString
 
     bool openResult = doOpenFiles(openInfos, openMineTypes);
 
-    bool openMount = doOpenFiles(mountOpenInfos, openMineTypes);
+    bool openMount = doOpenFiles(mountOpenInfos, mountOpenInfos);
 
     bool openCmd = doOpenFiles(cmdOpenInfos, cmdOpenInfos);
```
Johnson-zs: cmdOpenInfos 也没改吧？

liyigang1: 马上修改

max-lvs (approved):

PR #2130, opened 2024-07-19
fix: resolve the view jumping while click the same url

max-lvs:
verify the url is current or not

Log: resolve the view jumping while click the same url
Bug: https://pms.uniontech.com/bug-view-265031.html


Johnson-zs (changes requested):
tab会产生异常

PR #2135, opened 2024-07-22
fix: [preview] The music duration show 00:00

GongHeng2017:
When can not get duration, show "--"

Log: fix issue
Bug: https://pms.uniontech.com/bug-view-259569.html


src/plugins/common/dfmplugin-preview/pluginpreviews/music-preview/toolbarframe.cpp
```diff
@@ -130,6 +130,13 @@ void ToolBarFrame::stop()
 
 void ToolBarFrame::durationToLabel(qint64 duration)
 {
+    if (duration == 0) {
```
max-lvs: 这里最好用 duration < 1

max-lvs (approved):

PR #2136, opened 2024-07-22
fix: [dbus] use QProcess with arguments

itsXuSt:
as title.

Log: as title.


Johnson-zs (approved):

BLumia:
cf: https://github.com/linuxdeepin/developer-center/issues/3506

PR #2223, opened 2024-08-22
fix: Copy and paste the file to the root directory of FTP mount with an error message. After switching directories, an unknown file will be generated upon return

liyigang1:
Received FTP creation information, but FTP remote creation failed as the file does not exist. Modify to check if a file exists when adding a child

Log: Copy and paste the file to the root directory of FTP mount with an error message. After switching directories, an unknown file will be generated upon return
Bug: https://pms.uniontech.com/bug-view-269893.html

### DIFF
--- a/src/plugins/filemanager/core/dfmplugin-workspace/models/rootinfo.cpp
+++ b/src/plugins/filemanager/core/dfmplugin-workspace/models/rootinfo.cpp
@@ -9,6 +9,7 @@
 #include <dfm-base/utils/universalutils.h>
 #include <dfm-base/base/application/settings.h>
 #include <dfm-base/utils/fileutils.h>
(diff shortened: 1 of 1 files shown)


src/plugins/filemanager/core/dfmplugin-workspace/models/rootinfo.cpp
```diff
@@ -426,6 +427,8 @@ void RootInfo::addChildren(const QList<QUrl> &urlList)
 
     for (auto url : urlList) {
         url.setPath(url.path());
+        if (DeviceUtils::isFtp(url) && !DFMIO::DFile(url).exists())
```
max-lvs, line 430: gvfs ftp插件问题，非文管问题

max-lvs (commented):
-1

PR #2231, opened 2024-08-27
fix: dde-file-manager crash, clipboard crash

liyigang1:
dde-file-manager crash, clipboard crash

Log: dde-file-manager crash, clipboard crash


src/dfm-base/widgets/dfmwindow/filemanagerwindowsmanager.cpp
```diff
@@ -219,6 +219,7 @@ FileManagerWindowsManager::FMWindow *FileManagerWindowsManager::createWindow(con
         if (!(window->windowState() & Qt::WindowMaximized)) {
             window->moveCenter();
         }
+        emit firstWindowShow();
```
Johnson-zs: 此时window并没有被show

Johnson-zs (changes requested):

Johnson-zs (approved):

PR #2243, opened 2024-08-30
fix: When entering a file name or renaming, no prompt message is displayed after entering special characters

liyigang1:
The lack of a parent window in the tooltip resulted in loss of focus

Log: When entering a file name or renaming, no prompt message is displayed after entering special characters
Bug: https://pms.uniontech.com/bug-view-267147.html


max-lvs (dismissed):

src/plugins/filemanager/core/dfmplugin-workspace/views/iconitemeditor.cpp
```diff
@@ -353,6 +359,7 @@ bool IconItemEditor::eventFilter(QObject *obj, QEvent *ee)
         }
     }
     case QEvent::FocusOut:
+        qCritical() << obj << qApp->focusWidget() << d->edit;
```
Johnson-zs: 日志不规范，日志必须详细记录上下文信息，而不是只打印一个值。此外，使用fmCritical

Johnson-zs (changes requested):

Johnson-zs (approved):

PR #2302, opened 2024-09-20
fix: The application name on the dde-desktop cannot be renamed

liyigang1:
Desktop files on the desktop are hard linked and can be renamed based on whether the file is writable

Log: The application name on the dde-desktop cannot be renamed
Bug: https://pms.uniontech.com/bug-view-256967.html


Johnson-zs (changes requested):
abled？

Johnson-zs (approved):

PR #2325, opened 2024-10-09
fix: specify the version number of the partial dependency

wangling12:
Installing libdfm-extension-dev will only upgrade libdde-file-manager, but this cause file-manager or desktop to become unavailable. Specify version to solve similar problems.


wangling12:
话说, 看不出来哪没通过

Johnson-zs (approved):

PR #2393, opened 2024-11-11
fix: Fixed the crash issue of dragging files in desktop organization.

dengzhongyuan365-dev:

log: If the size bigger than list's size, use append instead.
Influence: Dragging files in desktop organization.

### DIFF
--- a/src/plugins/desktop/ddplugin-organizer/mode/collectiondataprovider.cpp
+++ b/src/plugins/desktop/ddplugin-organizer/mode/collectiondataprovider.cpp
@@ -94,7 +94,19 @@ void CollectionDataProvider::moveUrls(const QList<QUrl> &urls, const QString &ta
         // not belong collection
         return;
     }
-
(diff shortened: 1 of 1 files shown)


src/plugins/desktop/ddplugin-organizer/mode/collectiondataprovider.cpp
```diff
@@ -94,7 +94,19 @@ void CollectionDataProvider::moveUrls(const QList<QUrl> &urls, const QString &ta
         // not belong collection
         return;
     }
-
+    // In Qt6, when the size of the list is exceeded, the insert operation of QList will trigger an assertion crash
+    // To avoid the crash, if the size bigger than list's size, use append instead.
+    auto moveUrlInItems = [=](QHash<QString, CollectionBaseDataPtr>::Iterator &it, int targetIndex) {
```
Johnson-zs: 不建议用 = 捕获所有，需要什么直接写什么

PR #2396, opened 2024-11-11
fix: Fixed the crash issue of dragging files in desktop organization.

dengzhongyuan365-dev:

log: If the size bigger than list's size, use append instead.
Influence: Dragging files in desktop organization.


Johnson-zs (approved):

dengzhongyuan365-dev:
recheck

dengzhongyuan365-dev:
recheck

dengzhongyuan365-dev:
recheck

PR #2398, opened 2024-11-12
fix: Use the QList insert method to perform a validation check, and u…

dengzhongyuan365-dev:
…se append instead of insert when the index value is greater than the size of the list

The following are included: bookmark, tagdirmenu, workspace ..

log: The insert method of Qlist is compatible.

### DIFF
--- a/src/plugins/filemanager/core/dfmplugin-workspace/utils/filesortworker.cpp
+++ b/src/plugins/filemanager/core/dfmplugin-workspace/utils/filesortworker.cpp
@@ -20,6 +20,17 @@ using namespace dfmplugin_workspace;
 using namespace dfmbase::Global;
 using namespace dfmio;
 
+namespace {
+template<class T>
+void insertToList(QList<T> &list, int index, const T &t)
+{
+    if (index < 0 || index > list.size())
+        list.append(t);
+    else
+        list.insert(index, t);
+}
+}   // namespace
+
 FileSortWorker::FileSortWorker(const QUrl &url, const QString &key, FileViewFilterCallback callfun, const QStringList &nameFilters, const QDir::Filters filters, const QDirIterator::IteratorFlags flags, QObject *parent)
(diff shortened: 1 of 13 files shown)


src/plugins/filemanager/qml/dfmplugin-workspace/models/fileviewmodel.cpp
```diff
@@ -614,7 +614,13 @@ QList<ItemRoles> FileViewModel::getColumnRoles() const
         int customCount = roles.count();
         for (auto role : defualtColumnRoleList) {
             if (!roles.contains(role))
-                roles.insert(roles.length() - customCount, role);
+            {
```
Johnson-zs: 代码风格似乎没配置

src/plugins/filemanager/core/dfmplugin-workspace/utils/filesortworker.cpp
```diff
@@ -20,6 +20,17 @@ using namespace dfmplugin_workspace;
...
+namespace {
+template<class T>
+void insertToList(QList<T> &list, int index, const T &t)
+{
+    if (index < 0 || index > list.size())
+        list.append(t);
+    else
+        list.insert(index, t);
+}
+}
```
Johnson-zs: namespace后面需要加注释 // namespace XXX

src/plugins/filemanager/qml/dfmplugin-workspace/models/fileviewmodel.cpp
dengzhongyuan365-dev: 图片
已经使用插件对改动文件进行格式化，但是会出现异常格式化

src/plugins/filemanager/core/dfmplugin-workspace/utils/filesortworker.cpp
dengzhongyuan365-dev: 已经修改

Johnson-zs (approved):

PR #2408, opened 2024-11-14
fix: Fixed the issue that deleting a valut and creating a new one fai…

dengzhongyuan365-dev:
…led.

The process of creating the vault will start the cryfs process, and the cryfs process will occupy the files in the creation directory, resulting in the deletion of the vault not all the files are deleted, resulting in the failure to create the vault again in the original path.

log: During the deletion process, unmount the vaultLockPath and vaultUnlockPath first to prevent unsuccessful deletion.

bug: https://pms.uniontech.com/bug-view-284441.html
Influence: Creation and deletion of 'Vault'.

### DIFF
--- a/src/plugins/filemanager/dfmplugin-vault/utils/encryption/operatorcenter.cpp
+++ b/src/plugins/filemanager/dfmplugin-vault/utils/encryption/operatorcenter.cpp
@@ -6,6 +6,7 @@
 #include "vaultconfig.h"
 #include "utils/operator/pbkdf2.h"
 #include "utils/operator/rsam.h"
+#include "utils/pathmanager.h"
 
 #include <dfm-io/dfmio_utils.h>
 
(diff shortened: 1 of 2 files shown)


src/plugins/filemanager/dfmplugin-vault/utils/encryption/operatorcenter.cpp
```diff
@@ -21,6 +21,7 @@
 #include <QtGlobal>
 #include <QProcess>
 #include <QtConcurrent>
+#include "../pathmanager.h"
```
Johnson-zs: 编码规范，不得使用相对路径

dengzhongyuan365-dev: 图片
已经修改

PR #2470, opened 2024-12-06
fix: improve folder list index handling

wyu71:
- Store folder index in item's UserRole data instead of relying on row number
- Fix potential mismatch between visible items and data source
- Optimize variable initialization order

Log: improve folder list index handling
Bug: https://pms.uniontech.com/bug-view-289995.html


Johnson-zs (commented):
没关联PMS？

Johnson-zs (approved):

Johnson-zs (approved):

PR #2487, opened 2024-12-12
fix: improve splitter animation handling

Lighto-Ku:
1. Add handle enable/disable during splitter animation to prevent user interference
2. Add animation state tracking in titlebar to avoid unnecessary UI updates
3. Optimize placeholder width updates to prevent redundant resizing
4. Add animation start/end value tracking for better state management

Log: fix animation issue
Bug: https://pms.uniontech.com/bug-view-292639.html

### DIFF
--- a/src/dfm-base/widgets/dfmwindow/filemanagerwindow.cpp
+++ b/src/dfm-base/widgets/dfmwindow/filemanagerwindow.cpp
@@ -272,7 +272,10 @@ void FileManagerWindowPrivate::connectAnimationSignals()
         bool expanded = curSplitterAnimation->endValue().toInt() > 1;
         if (expanded)
             resetSideBarSize();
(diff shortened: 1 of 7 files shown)


src/dfm-base/widgets/dfmwindow/filemanagerwindow.cpp
```diff
@@ -272,6 +272,7 @@ void FileManagerWindowPrivate::connectAnimationSignals()
         bool expanded = curSplitterAnimation->endValue().toInt() > 1;
         if (expanded)
             resetSideBarSize();
+        splitter->handle(1)->setEnabled(expanded);
```
Johnson-zs: 这里的目的？另外最好要空指针检查

Johnson-zs (approved):

PR #2513, opened 2024-12-24
fix bus

dengzhongyuan365-dev:
fix: Fix the problem of deformation of the crumb menu.
    
    The rounded corner set by the system is too large. When it is displayed only one item, it cannot be completely drawn to the corner, causing the lower part of the window to abnormal. Compatible treatment will increase height when there is only one item.
    
    Log: Fix the problem of deformation of the crumb menu.
    
    Bug: https://pms.uniontech.com/bug-view-296751.html

### DIFF
--- a/src/plugins/filemanager/dfmplugin-titlebar/views/folderlistwidget.cpp
+++ b/src/plugins/filemanager/dfmplugin-titlebar/views/folderlistwidget.cpp
@@ -8,11 +8,11 @@
 
 #include <dfm-base/base/schemefactory.h>
 #include <dfm-base/base/application/application.h>
+#include <dfm-base/utils/chinese2pinyin.h>
 
 #include <QVBoxLayout>
 #include <QStandardItemModel>
 #include <QKeyEvent>
-#include <dfm-base/utils/chinese2pinyin.h>
 
 DWIDGET_USE_NAMESPACE
 using namespace dfmplugin_titlebar;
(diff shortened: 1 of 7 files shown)


src/plugins/filemanager/dfmplugin-titlebar/views/folderlistwidget.cpp
```diff
@@ -166,7 +166,13 @@ void FolderListWidget::setFolderList(const QList<CrumbData> &datas, bool stacked
...
 
     int folderCount = dataNum > kMaxFolderCount ? kMaxFolderCount : dataNum;
-    setFixedHeight(kFolderListItemMargin * 2 + kFolderItemHeight * folderCount);
+    if (dataNum > 1) {
+        d->folderView->setViewportMargins(kItemMargin, kItemMargin, kItemMargin, kItemMargin);
+        setFixedHeight(kItemMargin * 2 + kFolderItemHeight * folderCount);
+    } else {
+        d->folderView->setViewportMargins(kItemMargin, kItemMargin * 3 / 2, kItemMargin, kItemMargin * 3 / 2);
+        setFixedHeight(kItemMargin * 3 + kFolderItemHeight * folderCount);
+    }
```
Lighto-Ku, line 175: 这样修改不符合设计预期，只存在一条数据时，上下空白区域太大。

Lighto-Ku (changes requested):

src/plugins/filemanager/dfmplugin-titlebar/views/folderlistwidget.cpp
```diff
@@ -12,6 +12,7 @@
 #include <QVBoxLayout>
 #include <QStandardItemModel>
 #include <QKeyEvent>
+#include <dfm-base/utils/chinese2pinyin.h>
```
Johnson-zs: 头文件位置没放对

src/plugins/filemanager/dfmplugin-titlebar/views/folderlistwidget.cpp
dengzhongyuan365-dev: 这里和周倩讨论过，因为系统的圆角在只有一项的时候会绘制不全，出现残角。就还是加高让他显示正常。

src/plugins/filemanager/dfmplugin-titlebar/views/folderlistwidget.cpp
dengzhongyuan365-dev: 已经修改

Johnson-zs (approved):

PR #2573, opened 2025-02-12
fix: update path handling in isMptOfDevice method

Lighto-Ku:
- Changed the declaration of `path` from `const QString &` to `QString` to allow modification.
- Added a condition to set `path` to `/sysroot/` if it is equal to `/`.
- Removed the unused `bool ok` variable and simplified the condition to check if `urls` is not empty before assigning the first URL to `converted`.

Log: fix property display with wrong view
Bug: https://pms.uniontech.com/bug-view-299363.html

### DIFF
--- a/src/dfm-base/base/device/deviceproxymanager.cpp
+++ b/src/dfm-base/base/device/deviceproxymanager.cpp
@@ -147,10 +147,7 @@ bool DeviceProxyManager::isFileFromOptical(const QString &filePath)
 bool DeviceProxyManager::isMptOfDevice(const QString &filePath, QString &id)
 {
     d->initMounts();
-    QString path = filePath.endsWith("/") ? filePath : filePath + "/";
-    if (path == "/")
-        path = "/sysroot/";
-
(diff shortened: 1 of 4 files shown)


src/dfm-base/base/device/deviceproxymanager.cpp
```diff
@@ -147,7 +147,10 @@ bool DeviceProxyManager::isFileFromOptical(const QString &filePath)
 bool DeviceProxyManager::isMptOfDevice(const QString &filePath, QString &id)
 {
     d->initMounts();
-    const QString &path = filePath.endsWith("/") ? filePath : filePath + "/";
+    QString path = filePath.endsWith("/") ? filePath : filePath + "/";
+    if (path == "/")
+        path = "/sysroot/";
```
Johnson-zs: 这个sysroot没有依据

Johnson-zs (changes requested):

Johnson-zs (approved):

PR #2591, opened 2025-02-17
fix: Ensure plain text rendering in file preview and property dialogs

Johnson-zs:
- Set text format to plain text in UnknowFilePreview name label
- Use setPlainText instead of setText in KeyValueLabel and NameTextEdit
- Add plain text format to file name labels in EditStackedWidget

This change prevents potential HTML or rich text rendering issues in various file management UI components.

Log:

Bug: https://bbs.deepin.org/post/284207

### DIFF
--- a/src/plugins/common/dfmplugin-propertydialog/views/editstackedwidget.cpp
+++ b/src/plugins/common/dfmplugin-propertydialog/views/editstackedwidget.cpp
@@ -90,8 +90,8 @@ void NameTextEdit::slotTextChanged()
         text.chop(1);
     }
 
-    if (text.count() != old_text.count()) {
-        this->setText(text);
+    if (text.size() != old_text.size()) {
+        this->setPlainText(text);
     }
 
     QTextCursor cursor = this->textCursor();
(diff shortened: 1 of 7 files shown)


GongHeng2017 (approved):

src/plugins/common/dfmplugin-propertydialog/views/editstackedwidget.cpp
```diff
@@ -223,7 +223,7 @@ void EditStackedWidget::initUI()
 
 void EditStackedWidget::initTextShowFrame(QString fileName)
 {
-    auto originalFileName = fileName; // 存储原始文件名
+    auto originalFileName = fileName;   // 存储原始文件名
```
zccrs, line 226: 这行的变更没有意义？

zccrs (changes requested):

src/plugins/common/dfmplugin-propertydialog/views/editstackedwidget.cpp
Johnson-zs: 保存时启用了 .clang-format 自动整理了原来不合规的代码风格

PR #2632, opened 2025-02-27
fix: [file-dialog] optimize extension name search to prevent dialog sticking

GongHeng2017:
When searching for extension names in accept mode, the dialog would continue
iterating through all filters even after finding a match, causing potential
performance issues and UI sticking. Break the loop once a valid extension
is found.

Log: fix issue
Bug: https://pms.uniontech.com/bug-view-305621.html

## Summary by Sourcery

Bug Fixes:
- Fixes an issue where the file dialog could become stuck in accept mode by breaking the loop after finding a new extension.


Johnson-zs (changes requested):
stick 这个词不太规范，请使用AI总结

GongHeng2017:
> stick 这个词不太规范，请使用AI总结

done

Johnson-zs (approved):

PR #2723, opened 2025-03-26
fix: [filedialog] Open button can not clicked.

GongHeng2017:
--Previously, when receiving the QFileDialog::DirectoryOnly enum value from Qt5 code,
  it triggered incorrect logic in Qt6 where this enum no longer exists, This caused
  directory selection failures.

To resolve this:
1. Added interge comparison(static_cast<int>) instead of direct enum comparison
2. Maintained backward compatibility with Qt5 while working correctly in Qt6

Log: fix issue
Bug: https://pms.uniontech.com/bug-view-309043.html

## Summary by Sourcery

Bug Fixes:
- Fixes an issue where the Open button in the file dialog was not clickable when selecting directories in Qt6.

### DIFF
--- a/src/plugins/filedialog/core/views/filedialog.cpp
+++ b/src/plugins/filedialog/core/views/filedialog.cpp
@@ -45,6 +45,8 @@ Q_DECLARE_METATYPE(QDir::Filters);
 DFMBASE_USE_NAMESPACE
 using namespace filedialog_core;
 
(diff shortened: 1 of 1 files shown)


Johnson-zs (changes requested):
add pms url

Johnson-zs (approved):

PR #2939, opened 2025-06-22
show thumbnails for heif/heic images

SiamSami:
This PR implements support for generating thumbnails of heif/heic images using `heif-thumbnailer`


SiamSami:
will rewrite using libheif. Closing.

PR #3131, opened 2025-09-12
fix(filemanager): enhance window management and tab opening logic

Kakueeen:
- Improved the logic for opening new tabs and managing active windows in the file manager.
- Introduced a streamlined approach to activate existing windows or create new ones based on user settings.
- Updated signal dispatching for better responsiveness when opening new tabs or windows.

Log: These changes enhance the user experience by ensuring that the file manager effectively manages window states and tab interactions, leading to a more intuitive workflow.

## Summary by Sourcery

Refine the file manager’s window and tab management to use a consistent flag-based approach, activate windows on tab/window open, and update the related configuration key.

Enhancements:
- Streamline tab and window opening logic to respect the open-in-new-tab attribute and separate-process setting
- Automatically activate and restore (un-minimize) existing or newly created windows when opening tabs or windows
- Fallback to the last active or most recently created window when no active window ID is available

Chores:
- Bump the config accessor key for the open-folder-windows-in-a-separate-process setting from .01 to .02

### DIFF
--- a/src/apps/dde-file-manager/commandparser.cpp
+++ b/src/apps/dde-file-manager/commandparser.cpp
@@ -363,19 +363,30 @@ void CommandParser::openWindowWithUrl(const QUrl &url)
         });
     }
 
-    auto winId = FileManagerWindowsManager::instance().lastActivedWindowId();
-    if (Application::appAttribute(Application::kOpenInNewTab).toBool() && winId > 0) {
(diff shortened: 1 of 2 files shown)


Johnson-zs (commented):
没bug号？

Johnson-zs (approved):

PR #3448, opened 2025-12-23
feat: [services]add systemd service files and security hardening configs

re2zero:
Add systemd service files for disk encryption service and security hardening configs for mount control and share control services. The changes include:
- Create deepin-filemanager-diskencrypt.service systemd unit file
- Add security hardening configs with restricted access policies
- Update CMakeLists to install systemd service files
- Modify D-Bus service to use systemd service file
- Add systemd drop-in configurations for enhanced security

Log: add systemd service files and security hardening configs.

## Summary by Sourcery

Add systemd service integration and security hardening for file manager backend services.

New Features:
- Introduce a dedicated systemd service unit for the disk encryption service used by the file manager.

Enhancements:
- Add systemd drop-in security hardening configurations for the MountControl and UserShareManager services to tighten isolation and restrict access to sensitive paths.
- Update CMake installation rules to deploy the new systemd unit and drop-in configuration files alongside existing D-Bus service definitions.


debian/dde-file-manager-services-plugins.install
```diff
@@ -1,4 +1,5 @@
 usr/lib/*/deepin-service-manager/*.so
+usr/lib/systemd/system/
```
Johnson-zs: 写在filemanager的install

Johnson-zs (changes requested):

Johnson-zs (approved):

PR #3451, opened 2025-12-25
fix(delegate): add file icon fallback when thumbnail rendering fails

liyigang1:
When thumbnail pixmap is null or fails to render, the icon area remains empty. This change adds a fallback mechanism to display the file icon instead of showing a blank space.

Changes:
 - Add kItemFileIconRole to ItemRoles enum for accessing file icons
 - Modify paintIcon() to return bool indicating success/failure
 - Return QRect(-1, -1, -1, -1) when thumbnail pixmap is null
 - Add fallback logic in all delegates to use file icon when thumbnail fails
 - Update FileInfoModel and FileItemData to provide file icon via new role
 - Apply fallback in CanvasItemDelegate, CollectionItemDelegate, IconItemDelegate, BaseItemDelegate, ListItemPaintProxy, and TreeItemPaintProxy

This ensures users always see an icon representation even when thumbnails cannot be generated.

Log: add file icon fallback when thumbnail rendering fails
Bug: https://pms.uniontech.com/bug-view-340119.html

## Summary by Sourcery

Ensure file icons are always displayed in canvas, collection, icon, list, and tree views when thumbnail rendering fails by introducing a dedicated file icon role and thumbnail fallback handling across delegates and helpers.

Bug Fixes:
- Prevent empty icon areas when thumbnail pixmaps are null or cannot be rendered by falling back to the underlying file icon.

Enhancements:
- Introduce a kItemFileIconRole model role and plumb it through file models and item data to expose the raw file icon separately from thumbnails.
- Change shared icon painting helpers to report thumbnail rendering failures so delegates can trigger file icon fallback instead.

### DIFF
--- a/src/plugins/desktop/ddplugin-organizer/delegate/collectionitemdelegate.cpp
+++ b/src/plugins/desktop/ddplugin-organizer/delegate/collectionitemdelegate.cpp
@@ -157,12 +157,22 @@ void CollectionItemDelegate::paint(QPainter *painter, const QStyleOptionViewItem
         // draw icon and background
         const QRect rIcon = iconRect(option.rect);
         paintBackground(painter, indexOption, rIcon);
-        paintIcon(painter, indexOption.icon,
-                  { rIcon,
-                    Qt::AlignCenter,
-                    (option.state & QStyle::State_Enabled) ? QIcon::Normal : QIcon::Disabled,
-                    QIcon::Off,
-                    isThumnailIconIndex(index) });   // why Enabled?
+        const std::optional<QRect> &pIcon = paintIcon( painter, indexOption.icon,
+                                                       { rIcon,
+                                                         Qt::AlignCenter,
+                                                         (option.state & QStyle::State_Enabled) ? QIcon::Normal : QIcon::Disabled,
+                                                         QIcon::Off,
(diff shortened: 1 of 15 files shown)


src/plugins/desktop/ddplugin-organizer/delegate/collectionitemdelegate.cpp
```diff
@@ -778,6 +797,10 @@ QRect CollectionItemDelegate::paintIcon(QPainter *painter, const QIcon &icon, co
     const qreal pixelRatio = painter->device()->devicePixelRatioF();
     const QPixmap &px = getIconPixmap(icon, opts.rect.size().toSize(), pixelRatio, opts.mode, opts.state);
 
+    // 缩略图缩放到指定的size，绘制不出来就直接返回，绘制fileicon
+    if (px.isNull() && opts.isThumb)
```
Johnson-zs, line 805: std::optional

Johnson-zs (changes requested):

src/plugins/desktop/ddplugin-organizer/delegate/collectionitemdelegate.cpp
liyigang1: 已修改

Johnson-zs (approved):

PR #3601, opened 2026-02-27
fix: use linyaps's deepin-reader if deepin-reader is not available

SiamSami:
Deepin no longer ships with dpkg's version of deepin-reader. calling "deepin-reader" to generate thumbnails will not work. This patch fixes it by using linyaps's version of deepin-reader instead.


Johnson-zs (approved):
Thanks

PR #3679, opened 2026-03-24
master

Johnson-zs:
- **fix: optimize file rename selection and menu handling**
- **chore: bump version to 6.5.128**

## Summary by Sourcery

Optimize file rename result handling and context menu selection while updating version metadata.

Bug Fixes:
- Prevent auto-selecting renamed files when the count exceeds a safe limit or when there is already an active selection.
- Fall back to the current selection index when the focused file URL no longer maps to a valid model index during rename from the context menu.

Enhancements:
- Extract and reuse the delay calculation for selecting files after rename via a dedicated helper method.

Chores:
- Bump package version to 6.5.128.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,13 @@
+dde-file-manager (6.5.128) unstable; urgency=medium
+
+  * style: refactor elide text layout code formatting
+  * fix: correct notification parameter order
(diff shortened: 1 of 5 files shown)


Johnson-zs:
@sourcery-ai review

PR #3861, opened 2026-06-28
fix: correct desktop available geometry on treeland fractional scaling

JWWTSL:
log: Derive the true scale factor from the dock physical span divided by the logical screen edge, instead of QScreen::devicePixelRatio which on treeland fractional scaling returns a wrong wl_output integer scale (e.g. 2 for a real 1.25). This caused dock height to be under- subtracted, leaving the bottom row of desktop filenames hidden behind the dock.

PMS: BUG-366085

## Summary by Sourcery

Adjust desktop available geometry calculations to correctly account for dock size under Wayland/treeland fractional scaling.

Bug Fixes:
- Resolve incorrect available desktop area that caused bottom-row desktop filenames to be hidden behind the dock when using fractional scaling on treeland/Wayland.

Enhancements:
- Derive the effective scale factor from the dock’s physical span relative to the logical screen edge instead of relying solely on QScreen::devicePixelRatio.
- Improve dock containment checks to handle both physical and logical coordinate systems across X11 and Wayland.

### DIFF
--- a/src/plugins/desktop/ddplugin-core/screen/screenqt.cpp
+++ b/src/plugins/desktop/ddplugin-core/screen/screenqt.cpp
@@ -8,6 +8,7 @@
 #include <qpa/qplatformscreen.h>
 #include <QDebug>
 #include <QApplication>
+#include <QWindow>
 
 namespace GlobalPrivate {
 static QRect dealRectRatio(QRect orgRect)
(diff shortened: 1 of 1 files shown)


JWWTSL:
### ① ✅ 已采纳：回退比例改用本屏而非主屏 / Adopted: per-screen `devicePixelRatio`

已把回退值从 `qApp->primaryScreen()->devicePixelRatio()` 改为本屏 `qscreen->devicePixelRatio()`，
多屏不同分数缩放时不再误用主屏比例。（建议里的代码片段引用了本 PR 已删除的 `dealRectRatio`，
故只采纳其思路，未照抄。）

> Changed the fallback to this screen's own `qscreen->devicePixelRatio()` instead of the primary
> screen, so multi-monitor setups with different fractional scales no longer borrow the primary
> display's ratio. (The snippet referenced `dealRectRatio`, which this PR removes, so I applied the
> idea, not the snippet.) Already pushed.

### ② ❌ 不采纳（误报）："dock 未必铺满全屏边 → 少减 → 再次遮挡" / Not adopted (false positive)

这条的失效方向是反的，不会让遮挡复现：

- dock 是按 `dockH = dockrectI.height / ratio` 减的。ratio **偏小** → `dockH` **偏大** → 减得**更多**，不是更少。
- 反推的 ratio 有上界：dock 物理跨度不可能超过屏幕物理宽，即 `dockrectI.width / ret.width ≤ 真实缩放`；
  再叠加 `ratio = max(ratio, 1.0)` 的钳制，算出的 `dockH` **恒 ≥ 正确值** → 只会**多减**（底部留几像素白），
  绝不会**少减**（遮挡）。
- 反过来，按建议在"非铺满"时回退到 `devicePixelRatio()`，在 treeland 上恰恰会取到错误的整数值
  （真实 1.25 被上报成 2）→ `dockH = 高 / 2` → **少减 → 遮挡复现**。避开这个错值正是本 PR 的目的。

所以"窄/居中 dock"场景下当前实现是安全的，最坏只是底部多几像素留白，不会遮挡文件名。

> The failure direction is inverted: `dockH = height / ratio`, so a smaller ratio subtracts MORE, not
> less. The derived ratio is bounded (`dock span ≤ screen physical width`) and clamped to `≥ 1`, so
> `dockH` is always `≥` the correct value → only ever over-subtract (a few px of bottom margin), never
> under-subtract. Conversely, the suggested fallback to `devicePixelRatio()` would reintroduce the bug
> on treeland (it reports 2 for a real 1.25 → `height / 2` → occlusion again) — avoiding that value is
> the whole point of this PR.

### ③ 🟡 暂缓：抽取 `effectiveDockScaleFactor` / `dockBelongsToScreen` helper / Deferred: extract helpers

合理的可读性改进，不改行为。为保持本次修复最小、聚焦 bug（已在 treeland 125% 验证），
计划放到后续单独重构里做。

> Fair readability improvement, no behavior change. Keeping this PR minimal and focused on the bug
> (verified on treeland @ 125%); will do the extraction in a follow-up.

### ④ 🟡 暂缓：dock 位置 magic number `0/2` 换具名常量 / Deferred: named dock-position constants

同样合理。但函数内的 `switch (dockPos)` 也用了同样的裸数字，要改应统一替换，故一并放到后续重构。

> Also reasonable, but the `switch (dockPos)` below uses the same raw numbers; will replace them
> together in the same follow-up for consistency.

Johnson-zs (approved):

PR #3923, opened 2026-07-13
Feat(filedialog): improve address bar navigation and Enter key handli…

GongHeng2017:
…ng in chooser mode

Add a new utility `UniversalUtils::isChooserDialogProcess()` to detect whether the current process is a file chooser / select dialog (dde-file-dialog, dde-select-dialog-x11, dde-select-dialog-wayland).

Two related fixes for the file chooser dialog:

1. In `FileDialog::handleEnterInSaveMode()`, return early when the focused widget is inside the title bar (e.g. the address bar). The Enter key should trigger address bar navigation, not the save action.

2. In `TitleBarHelper::handleJumpToPressed()`, when running as a chooser dialog process, jumping to a file path in the address bar now navigates to the file's parent directory and selects the file (via a `selectUrl` query parameter) instead of opening it directly, which matches the expected chooser dialog behavior.

Log: Add feature for chooser dialog.
Task: https://pms.uniontech.com/task-view-392211.html

## Summary by Sourcery

Improve file chooser dialog behavior for address bar navigation and Enter key handling in save mode.

New Features:
- Add a utility to detect whether the current process is running as a file chooser or select dialog.

Bug Fixes:
- Prevent Enter key presses in the title bar address bar from triggering the save action in save mode.
- Ensure jumping to a file path in the address bar in chooser dialogs navigates to the parent directory and selects the file instead of opening it directly.

### DIFF
--- a/src/plugins/filedialog/core/core.cpp
+++ b/src/plugins/filedialog/core/core.cpp
@@ -13,6 +13,10 @@
 #include <dfm-base/widgets/filemanagerwindowsmanager.h>
 #include <dfm-base/base/application/application.h>
 #include <dfm-base/base/application/settings.h>
+#include <dfm-base/dfm_event_defines.h>
+#include <dfm-base/base/schemefactory.h>
+
+#include <QUrlQuery>
 
 #include <QDBusError>
 #include <QDBusConnection>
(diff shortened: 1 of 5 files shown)


src/dfm-base/utils/universalutils.cpp
```diff
@@ -648,6 +648,24 @@ void UniversalUtils::setDockDnDMimeData(QMimeData *mimeData, const QUrl &url, co
...
 }
 
+bool UniversalUtils::isChooserDialogProcess()
+{
+    static bool result = false;
+
+    static bool hasCall = false;
+    if (!hasCall) {
+        const QString appName = qAppName();
+        if (appName == "dde-file-dialog"
```
Johnson-zs: 不能写这种死代码

Johnson-zs (changes requested):

src/dfm-base/utils/universalutils.cpp
GongHeng2017: done

Johnson-zs (approved):

PR #4249, opened 2026-08-26
Index service resource optimization.

Johnson-zs:
- **feat: implement intelligent task scheduling with environmental awareness**
- **feat: add environment-aware task scheduling with priority queuing**
- **refactor: extract task launching and queuing logic**
- **refactor: refine CPU quota management with reference counting**
- **feat: refine index task status reporting with detailed blocking conditions**
- **feat: implement server-driven index status management**
- **fix: remove silent parameter from index task calls**
- **feat: add comprehensive unit tests for TaskManager grading logic**
- **feat: add force bypass for critical file tasks**
- **feat: refine index status logic with environmental checks**
- **feat: enhance indexing control with manual and bypass options**

## Summary by Sourcery

Optimize text and OCR indexing by scheduling work according to task cost and device conditions while providing actionable, server-driven status updates.

New Features:
- Add environment-aware task grading, priority scheduling, queueing, pausing, and force-bypass controls for text and OCR indexing.
- Expose server-driven index status through D-Bus and present detailed waiting states with manual or environment-bypass actions in the file manager.
- Add configurable workload thresholds for classifying incremental indexing tasks.

Bug Fixes:
- Prevent concurrent indexing tasks from incorrectly overriding shared CPU quota settings through reference-counted resource management.
- Remove obsolete silent-task parameters and centralize CPU resource policy cleanup around task execution.

Enhancements:
- Refactor task launching and scheduling around graded tasks, environment changes, resumable pauses, duplicate suppression, and authoritative status reporting.
- Treat file deletion and move operations as lightweight synchronization tasks that bypass environmental restrictions.

Tests:
- Expand unit coverage for task grades, environment decision matrices, workload thresholds, status generation, priority ordering, and pause/resource-policy behavior.

Chores:
- Convert EnvDetector to a shared singleton used by indexing runtimes and monitoring tools.

### DIFF
--- a/src/services/textindex/task/taskhandler.cpp
+++ b/src/services/textindex/task/taskhandler.cpp
@@ -485,7 +485,7 @@ bool cleanupIndexs(const IndexContext &context, IndexReaderPtr reader, IndexWrit
         int removedCount = 0;
 
         // 检查每个文档对应的文件是否存在
-        for (int32_t i = 0; i < allDocs->totalHits && running.isRunning(); ++i) {
+        for (int32_t i = 0; i < allDocs->totalHits && running.isRunning() && !running.isPauseRequested(); ++i) {
             // Ensure scoreDocs[i] is not null before accessing ->doc
             if (!allDocs->scoreDocs || !allDocs->scoreDocs[i]) {
                 fmWarning() << "[cleanupIndexs] Null scoreDoc at index" << i << "during index cleanup";
@@ -738,6 +738,14 @@ TaskHandler TaskHandlers::CreateIndexHandler(const IndexContext &context)
                       migrator.isActive() ? &migrator : nullptr);
             });
 
+            if (running.isPauseRequested()) {
+                fmInfo() << "[CreateIndexHandler] Index creation paused by scheduler request";
+                writer->commit();
+                result.paused = true;
+                result.success = false;
+                return result;
+            }
+
             // Only the creation of an index that is interrupted is also considered a failure
             // Created indexes must be guaranteed to be complete
             if (!running.isRunning()) {
@@ -857,6 +865,14 @@ TaskHandler TaskHandlers::UpdateIndexHandler(const IndexContext &context)
                        migrator.isActive() ? &migrator : nullptr);
             });
 
+            if (running.isPauseRequested()) {
+                fmInfo() << "[UpdateIndexHandler] Index update paused by scheduler request";
+                writer->commit();
+                result.paused = true;
+                result.success = false;
+                return result;
+            }
+
             if (!running.isRunning()) {
                 fmWarning() << "[UpdateIndexHandler] Index update was interrupted by user request";
                 result.interrupted = true;
@@ -956,6 +972,15 @@ TaskHandler TaskHandlers::CreateOrUpdateFileListHandler(const IndexContext &cont
             provider->traverse(running, [&](const QString &file) {
                 updateFile(context, file, excludeMatcher, reader, writer, &reporter);
             });
+
+            if (running.isPauseRequested()) {
+                fmInfo() << "[CreateOrUpdateFileListHandler] File list update paused by scheduler request";
+                writer->commit();
+                result.paused = true;
+                result.success = false;
+                result.remainingFiles = fileList;
+                return result;
+            }
 
             if (!running.isRunning()) {
                 fmWarning() << "[CreateOrUpdateFileListHandler] File list update was interrupted by user request";
--- a/src/services/textindex/task/taskhandler.cpp
+++ b/src/services/textindex/task/taskhandler.cpp
@@ -198,7 +198,7 @@ private:
 using FileHandler = std::function<void(const QString &path)>;
 
 DocumentPtr createFileDocument(const IndexContext &context, const QString &file,
-                                  const IndexContentMigrator *migrator = nullptr)
+                               const IndexContentMigrator *migrator = nullptr)
 {
     try {
         if (!context.extractor() || !context.documentBuilder()) {
@@ -735,7 +735,7 @@ TaskHandler TaskHandlers::CreateIndexHandler(const IndexContext &context)
 
             provider->traverse(running, [&](const QString &file) {
                 processFile(context, file, excludeMatcher, writer, &reporter,
-                      migrator.isActive() ? &migrator : nullptr);
+                            migrator.isActive() ? &migrator : nullptr);
             });
 
             if (running.isPauseRequested()) {
@@ -843,6 +843,14 @@ TaskHandler TaskHandlers::UpdateIndexHandler(const IndexContext &context)
                 return result;
             }
 
+            if (running.isPauseRequested()) {
+                fmInfo() << "[UpdateIndexHandler] Index update paused during index cleanup";
+                writer->commit();
+                result.paused = true;
+                result.success = false;
+                return result;
+            }
+
             // 使用文件提供者遍历文件
             auto provider = createFileProvider(context, path);
             if (!provider) {
@@ -862,7 +870,7 @@ TaskHandler TaskHandlers::UpdateIndexHandler(const IndexContext &context)
 
             provider->traverse(running, [&](const QString &file) {
                 updateFile(context, file, excludeMatcher, reader, writer, &reporter,
-                       migrator.isActive() ? &migrator : nullptr);
+                           migrator.isActive() ? &migrator : nullptr);
             });
 
             if (running.isPauseRequested()) {
--- a/src/services/textindex/task/taskmanager.cpp
+++ b/src/services/textindex/task/taskmanager.cpp
@@ -638,14 +638,46 @@ std::optional<IndexTask::Grade> TaskManager::currentTaskGrade() const
 
 QString TaskManager::currentIndexStatus() const
 {
-    if (!hasRunningTask() && !currentTask) {
-        if (!taskQueue.isEmpty())
-            return "Blocked";
+    if (hasRunningTask() && currentTask)
+        return "Running";
+
+    if (!hasRunningTask() && !currentTask && !taskQueue.isEmpty()) {
+        // Tasks queued but not running → blocked by environment.
+        const EnvState env = EnvDetector::instance().currentState();
+        const IndexTask::Grade headGrade = taskQueue.head().grade;
+
+        if (canRun(headGrade, taskQueue.head().forceBypass, env))
+            return "Running";
+
+        // Distinguish by the first unsatisfied condition.
+        // The order matters: battery is the most restrictive, then power-save, then idle.
+        // Light tasks only require !powerSaveMode, so they won't trigger WaitingPower/WaitingIdle.
+        if (env.onBattery && headGrade != IndexTask::Grade::Light)
+            return "WaitingPower";
+        if (env.powerSaveMode)
+            return "WaitingPowerSave";
+        if (!env.idle && headGrade != IndexTask::Grade::Light)
+            return "WaitingIdle";
+
+        // Should not reach here: canRun returned false but no known condition matched.
+        // This can only happen with an unknown grade (None) — treat as Idle.
+        fmWarning() << "[TaskManager::currentIndexStatus] Unreachable: canRun=false but no condition matched"
+                    << "grade:" << static_cast<int>(headGrade)
+                    << "battery:" << env.onBattery
+                    << "powerSave:" << env.powerSaveMode
+                    << "idle:" << env.idle;
         return "Idle";
     }
 
-    if (hasRunningTask() && currentTask)
-        return "Running";
+    // No running task, no queued tasks
+    if (m_lastTaskFailed)
+        return "Failed";
+
+    // Idle: index exists → Completed; index not yet built → WaitingUpgrade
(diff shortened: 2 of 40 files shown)


Johnson-zs (commented):
## Review: Index service resource optimization

Thanks for this well-structured PR. The environment-aware task grading system is a solid design — the grade×environment matrix, cooperative pausing, reference-counted CPU quota, and server-driven status reporting are all clean approaches. The test coverage for the grading matrix is particularly thorough.

CI passes (api-check, cppcheck, static-check, debian-check all green). Below are the findings I'd like to flag, organized by severity.

---

### Bugs / correctness issues

**1. `shouldPreempt()` is dead code — preemption is tested but never wired in**

`shouldPreempt()` is defined and has 10 unit tests, but it is never called anywhere in the scheduling path. When a new `startTask` arrives while another is running, the code unconditionally calls `stopCurrentTask()` and clears the queue — no grade comparison is performed. This means a low-grade auto-update (Light) arriving while a Heavy Create is running will preempt it, which contradicts the priority semantics that `shouldPreempt` was designed to enforce. Either wire `shouldPreempt()` into `startTask()` (only stop the current task when `shouldPreempt(newGrade, currentGrade)` is true) or remove the function and its tests to avoid implying a feature that doesn't exist.

**2. `onTaskPaused()` loses `forceBypass` and non-primary paths on resume**

When a task is paused, the re-queue logic constructs a new `TaskQueueItem` but:
- **Multi-path loss**: `taskPath` is only the *primary* path (`pathList.first()`). The full `pathList` is captured in the handler lambda but not accessible from `IndexTask::taskPath()`. On resume, only the primary path is recovered — additional paths are silently dropped and never re-indexed.
- **`forceBypass` loss**: The new `TaskQueueItem` defaults to `forceBypass = false`. A task started via `ForceUpdateIndex(manual=true)` (which sets `forceBypass=true`) will, after being paused by an env change, be re-queued without `forceBypass`. It then won't resume until the environment is fully suitable, even though the user explicitly requested it.

Consider storing the full `pathList` and `forceBypass` on the `IndexTask` object so they can be recovered in `onTaskPaused()`.

**3. `onTaskFinished()` emits grade `"none"` even when tasks are queued**

```cpp
emit indexStatusChanged(currentIndexStatus(), gradeToString(IndexTask::Grade::None));
```
At this point `currentTask` is null (cleaned up), but the queue may contain blocked tasks. `currentIndexStatus()` correctly returns a waiting state (e.g. `"WaitingPower"`), but the grade is hardcoded to `"none"` instead of the queue head's grade. If `schedule()` doesn't start a new task (all blocked), clients receive `"WaitingPower"` + grade `"none"`, which is misleading. Use `gradeToString(currentOrQueuedGrade().value_or(IndexTask::Grade::None))` for consistency (as is already done in `onEnvStateChanged()` and `schedule()`).

**4. `startQueuedTask()` discards grade and `forceBypass` for file-list tasks**

```cpp
startFileListTask(item.type, item.fileList);
```
This passes only `type` and `fileList`, ignoring `item.grade` and `item.forceBypass`. Inside `startFileListTask()`, the grade is re-computed via `gradeFileListTask()` (which does synchronous `stat()` on every file) and `forceBypass` is recalculated as `type == RemoveFileList`. This means:
- The grade stored in the queue item (potentially set by `onTaskPaused()`) is lost.
- `forceBypass` is lost — a paused UpdateFileList task resumed from the queue will be re-graded and subject to environment checks it was previously exempt from.
- Every scheduling attempt re-stats all files, wasting CPU.

Consider adding an overload of `startFileListTask()` that accepts `grade` and `forceBypass`, or pass them through `startQueuedTask()`.

---

### Design / performance concerns

**5. `gradeFileListTask()` does synchronous `stat()` on the calling (main/DBus) thread**

`QFileInfo::size()` issues a `stat()` syscall per file. For large file-change bursts (potentially thousands of files), this blocks the DBus/main thread and could cause DBus timeouts. Consider caching the grade on the `TaskQueueItem` (set once when the task is first created) and reusing it in `startQueuedTask()` instead of re-grading.

**6. Paused file-list tasks re-process the entire file list on resume**

In `CreateOrUpdateFileListHandler`, when a pause is detected: `result.remainingFiles = fileList;` returns the *full* original file list, not the subset that wasn't yet processed. On resume, all files are re-indexed. For OCR tasks with thousands of images, this wastes significant CPU. The `traverse()` callback could track the last-processed index and return only the unprocessed tail.

---

### Minor issues / nits

**7. `releaseCpuQuota()` can decrement `g_limitedTaskCount` below zero** — no guard against unbalanced acquire/release. The current code paths are balanced, but a future bug could silently corrupt the counter. Consider `if (g_limitedTaskCount > 0) --g_limitedTaskCount;`.

**8. `IndexTask::Status::Ready` and `Status::Blocked` enum values are declared but never assigned** — dead enum values. Either use them or remove them.

**9. Test `createTempFile()` writes full bytes instead of sparse files** — `f->write(QByteArray(sizeBytes, 'x'))` writes the entire 200+ MB to disk. Four test cases create ~200 MB files (~800 MB total I/O), which is slow in CI. A true sparse file (`fseek` to `sizeBytes` + write 1 byte) would be much faster and still report the correct size via `stat()`.

---

### What is done well
- Clean refactoring of task launching into `launchTask()` / `tryEnqueueIfBlocked()` / `schedule()`
- Reference-counted CPU quota with mutex protection is the right approach for the shared-service model
- The `currentIndexStatus()` state machine is well-documented with clear comments explaining the Heavy/Light display rules
- DBus changes are additive (new methods/signals), backward compatible
- Comprehensive grading-matrix tests covering all grade×environment combinations

### Summary

The architecture is sound and the PR is close to mergeable. I'd recommend addressing **#1** (dead `shouldPreempt` code) and **#2** (pause/resume data loss) before merge, as they represent correctness gaps between the documented/tested behavior and the actual implementation. The remaining items can be addressed as follow-ups.

src/services/textindex/task/taskmanager.cpp
```diff
@@ -426,6 +440,330 @@ void TaskManager::onTaskProgress(IndexTask::Type type, qint64 count, qint64 tota
...
+        return !env.powerSaveMode;
+    case IndexTask::Grade::Medium:
+    case IndexTask::Grade::Heavy:
+        return !env.onBattery && !env.powerSaveMode && env.idle;
+    default:
+        return false;
+    }
+}
+
+bool TaskManager::shouldPreempt(IndexTask::Grade newGrade, IndexTask::Grade currentGrade) const
```
Johnson-zs: **Dead code — `shouldPreempt()` is never called.**

This function is defined and has 10 unit tests, but it is not invoked anywhere in the scheduling path. When a new `startTask` arrives while another is running, the code unconditionally calls `stopCurrentTask()` and clears the queue — no grade comparison is performed. This means a Light-grade auto-update can preempt a Heavy Create, contradicting the priority semantics `shouldPreempt` was designed to enforce.

Either wire this into `startTask()` (only stop the current task when `shouldPreempt(newGrade, currentGrade)` returns true) or remove the function and its tests to avoid implying a feature that doesn't exist.

src/services/textindex/task/taskmanager.cpp
```diff
@@ -426,6 +440,330 @@ void TaskManager::onTaskProgress(IndexTask::Type type, qint64 count, qint64 tota
...
+    if (!result.remainingFiles.isEmpty()) {
+        // File list task: save remaining files
+        item.type = IndexTask::Type::UpdateFileList;
+        item.fileList = result.remainingFiles;
+        item.path = QString("Resumed-%1").arg(QDateTime::currentDateTime().toString("yyyyMMdd-hhmmss"));
+    } else {
+        // Full-scan task (Heavy or Light/Medium): resume as Update (skip already indexed files)
+        item.type = IndexTask::Type::Update;
+        item.path = taskPath;
+        item.pathList = QStringList { taskPath };
```
Johnson-zs, line 468: **Pause/resume loses `forceBypass` and non-primary paths.**

Two issues with this re-queue logic:

1. **Multi-path loss**: `taskPath` is only the *primary* path (`pathList.first()`). If the original task was started with multiple paths, the remaining paths are silently dropped and never re-indexed after resume. The full `pathList` is captured in the handler lambda but is not accessible from `IndexTask::taskPath()`.

2. **`forceBypass` loss**: The new `TaskQueueItem` defaults to `forceBypass = false`. A task started via `ForceUpdateIndex(manual=true)` (forceBypass=true) that gets paused by an env change will be re-queued without `forceBypass`, so it won't resume until the environment is fully suitable — even though the user explicitly requested the update.

Consider storing the full `pathList` and `forceBypass` on the `IndexTask` object so they can be recovered here.

src/services/textindex/task/taskmanager.cpp
```diff
@@ -447,13 +785,19 @@ void TaskManager::onTaskFinished(IndexTask::Type type, HandlerResult result)
...
+        m_lastTaskFailed = !result.success;
+
     emit taskFinished(typeToString(type), taskPath, result.success);
     cleanupTask();
 
-    if (startNextTask()) {
-        fmInfo() << "[TaskManager::onTaskFinished] Started next queued task";
-    } else {
-        fmDebug() << "[TaskManager::onTaskFinished] No more tasks in queue";
+    emit indexStatusChanged(currentIndexStatus(), gradeToString(IndexTask::Grade::None));
```
Johnson-zs, line 784: **Grade hardcoded to `"none"` when tasks may be queued.**

At this point `currentTask` is null (cleaned up), but the queue may contain blocked tasks. `currentIndexStatus()` correctly returns a waiting state, but the grade is always `"none"`. If `schedule()` doesn't start a new task (all blocked by environment), clients receive e.g. `"WaitingPower"` + grade `"none"` instead of the queue head's actual grade.

Use `gradeToString(currentOrQueuedGrade().value_or(IndexTask::Grade::None))` for consistency — this is already what `onEnvStateChanged()` and `schedule()` do.

src/services/textindex/task/taskmanager.cpp
```diff
@@ -426,6 +440,330 @@ void TaskManager::onTaskProgress(IndexTask::Type type, qint64 count, qint64 tota
...
+    TaskQueueItem item = taskQueue.takeAt(bestIdx);
+    fmInfo() << "[TaskManager::schedule] Scheduling task - type:" << static_cast<int>(item.type)
+             << "grade:" << static_cast<int>(item.grade) << "remaining:" << taskQueue.count();
+    startQueuedTask(item);
+}
+
+void TaskManager::startQueuedTask(const TaskQueueItem &item)
+{
+    if (item.type == IndexTask::Type::CreateFileList
+        || item.type == IndexTask::Type::UpdateFileList
```
Johnson-zs, line 626: **Re-grading discards queue item grade and `forceBypass`.**

`startFileListTask(item.type, item.fileList)` passes only `type` and `fileList`, ignoring `item.grade` and `item.forceBypass`. Inside `startFileListTask()`, the grade is re-computed via `gradeFileListTask()` (synchronous `stat()` on every file) and `forceBypass` is recalculated as `type == RemoveFileList`.

This means: (1) the grade set by `onTaskPaused()` is lost, (2) `forceBypass` is lost — a paused UpdateFileList task resumed from the queue gets re-graded and subject to env checks it was previously exempt from, (3) every scheduling attempt re-stats all files.

Consider passing `item.grade` and `item.forceBypass` through to `startFileListTask()`.

src/services/textindex/task/taskmanager.cpp
```diff
@@ -609,57 +959,74 @@ void TaskManager::cleanupTask()
...
+            : config.lightIncrementFileCountThreshold();
+
+    if (fileList.size() > countThreshold)
+        return IndexTask::Grade::Medium;
+
+    qint64 sizeThreshold = config.lightIncrementSizeThresholdMB() * 1024 * 1024;
+    qint64 totalSize = 0;
+    for (const auto &filePath : fileList) {
+        QFileInfo info(filePath);
+        totalSize += info.size();
```
Johnson-zs, line 993: **Synchronous `stat()` on the main/DBus thread.**

`QFileInfo::size()` issues a `stat()` syscall per file. This runs on the DBus/main thread and could block for thousands of files in a large file-change burst. Consider caching the grade on the `TaskQueueItem` when the task is first created and reusing it, rather than re-grading on every scheduling attempt.

src/services/textindex/task/taskhandler.cpp
```diff
@@ -957,6 +981,15 @@ TaskHandler TaskHandlers::CreateOrUpdateFileListHandler(const IndexContext &cont
                 updateFile(context, file, excludeMatcher, reader, writer, &reporter);
             });
 
+            if (running.isPauseRequested()) {
+                fmInfo() << "[CreateOrUpdateFileListHandler] File list update paused by scheduler request";
+                writer->commit();
+                result.paused = true;
+                result.success = false;
+                result.remainingFiles = fileList;
```
Johnson-zs, line 989: **Returns the full file list, not just unprocessed files.**

`result.remainingFiles = fileList` returns the *entire* original `fileList`. On resume, all files are re-indexed, even those already processed before the pause. For OCR tasks with thousands of images, this wastes significant CPU. The `traverse()` callback could track the last-processed index and return only the unprocessed tail.

PR #4250, opened 2026-08-26
feat: add view hint live-update and side widgets

Kakueeen:
1. Extend ViewHintWidget to support left/right custom widget slots via
Side enum
2. Add ViewHintCustomWidgetFactory for deferred widget creation on both
sides
3. Implement live-update in ViewHintMessage for icon, text, actions and
custom widgets after show
4. Add refresh() slot and closed() signal to ViewHintMessage
5. Extract rebuildRightWidget() to combine right custom widget with
action buttons
6. Add slot_ShowViewHint/slot_CloseViewHint/slot_UpdateViewHint to
workspace plugin
7. Extend ViewHintSpec with leftCustomWidgetFactory and
rightCustomWidgetFactory fields
8. Refactor search hint logic from SearchHelper to dedicated
SearchHintController
9. Add comprehensive unit tests for live-update and side widget
scenarios

Log: Enhanced view hint system with left/right custom widget support and
live content updates

Influence:
1. Test view hint display with left custom widget before icon
2. Test view hint display with right custom widget replacing action
buttons area
3. Test live-update of icon, text and actions after hint is already
visible
4. Test setting custom widget factory before and after show() is called
5. Test null factory assignment to clear existing custom widgets
6. Test refresh() method rebuilds all dynamic content correctly
7. Test search authorization hint still appears and functions correctly
8. Verify closed() signal is emitted when message is dismissed
9. Test concurrent left and right custom widget coexistence

feat: 添加视图提示实时更新和侧边自定义控件功能

1. 扩展 ViewHintWidget 支持通过 Side 枚举设置左侧/右侧自定义控件
2. 添加 ViewHintCustomWidgetFactory 支持两侧控件的延迟创建
3. 实现 ViewHintMessage 在显示后实时更新图标、文本、操作按钮和自定义控件
4. 为 ViewHintMessage 添加 refresh() 槽函数和 closed() 信号
5. 提取 rebuildRightWidget() 方法合并右侧自定义控件和操作按钮
6. 向 workspace 插件添加 slot_ShowViewHint/slot_CloseViewHint/
slot_UpdateViewHint 接口
7. 扩展 ViewHintSpec 添加 leftCustomWidgetFactory 和
rightCustomWidgetFactory 字段
8. 将搜索提示逻辑从 SearchHelper 重构到独立的 SearchHintController 类
9. 添加针对实时更新和侧边控件场景的全面单元测试

Log: 增强视图提示系统，支持左右自定义控件及内容实时更新

Influence:
1. 测试在图标前显示左侧自定义控件的视图提示
2. 测试在操作按钮区域显示右侧自定义控件
3. 测试提示可见后实时更新图标、文本和操作按钮
4. 测试在 show() 调用前后设置自定义控件工厂
5. 测试设置空工厂以清除现有自定义控件
6. 测试 refresh() 方法正确重建所有动态内容
7. 测试搜索授权提示仍能正常显示和功能正常
8. 验证消息关闭时发出 closed() 信号
9. 测试左侧和右侧自定义控件同时共存

## Summary by Sourcery

Enhance view hints with live content updates, extensible side widgets, and workspace-driven lifecycle control.

New Features:
- Add left and right custom widget slots with deferred factories to view hints.
- Support showing, closing, and updating view hints through workspace plugin interfaces.

Bug Fixes:
- Keep visible hints synchronized when icons, text, actions, or custom widgets change.
- Prevent stale custom widgets and accumulated layout spacers when replacing or clearing side widgets.
- Emit a closed signal when a view hint is dismissed while safely handling asynchronous destruction.

Enhancements:
- Refactor search authorization hint behavior into a dedicated SearchHintController.
- Allow view hint content and side widgets to be updated dynamically after display.

Tests:
- Add unit coverage for live updates, side widget combinations, factory replacement and clearing, refresh behavior, layout stability, and asynchronous closure.


Kakueeen (commented):
## Review: feat: add view hint live-update and side widgets

Thanks for the PR — the direction (deferred left/right widget factories, live updates, extracting `SearchHintController`) is sound, and the new unit coverage is welcome. CI/build/cppcheck/static-check all pass. Below are findings by severity. The two top items are real correctness bugs that I'd want addressed before merge; the rest are design/nit suggestions.

---

### 🔴 Bug 1 — Right-side spacer accumulates on every rebuild (`viewhintwidget.cpp`)

`ViewHintWidget::setCustomWidget(widget, Side::Right)` always inserts a fresh spacer at index 2, but when replacing an existing right widget it only removes the *widget*, never the spacer that accompanied the previous one:

```cpp
if (slot) {
    m_layout->removeWidget(slot);   // removes widget, leaves its spacer behind
    slot->deleteLater();
}
slot = widget;
...
    m_layout->insertSpacing(insertIdx, kCustomWidgetGap - kItemSpacing);  // new spacer every call
    m_layout->insertWidget(insertIdx + 1, widget);
```

This is reachable on every live update path: `ViewHintMessage::setActions()`, `setCustomWidgetFactory(_, Right)` and `refresh()` all funnel into `rebuildRightWidget()` → `setCustomWidget(_, Right)`. Each rebuild leaks one spacer, so the close button drifts right and an extra 40px gap appears per update. With repeated updates the layout degrades visibly.

Suggested fix — track/remove the spacer alongside the widget, or rebuild the right region instead of inserting a new spacer each time:

```cpp
if (slot) {
    const int idx = m_layout->indexOf(slot);
    m_layout->removeWidget(slot);
    if (side == Side::Right && idx > 0)
        delete m_layout->takeAt(idx - 1);   // remove the paired spacer
    slot->deleteLater();
}
```

(Note: the existing unit tests only assert `EXPECT_NO_FATAL_FAILURE`, so this leak is not caught — see Nits.)

---

### 🔴 Bug 2 — `applyContentToHint` resets unspecified fields to spec defaults on live updates (`workspacepage.cpp`)

`handleViewHintUpdate()` calls `applyContentToHint(currentHint, updates, spec)` for every update. But `applyContentToHint` falls back to *spec defaults* for any key absent from `updates` whenever `content` is non-empty:

```cpp
if (content.contains(kIcon))        hint->setIcon(...from content...);
else if (!spec.icon.isEmpty())      hint->setIcon(spec.icon);          // resets to spec

if (content.contains(kActions))     { ...from content... }
else if (!content.isEmpty())        hint->setActions(spec.actions);    // resets to spec

// left/right factory: else-if spec -> setCustomWidgetFactory(spec...)  // re-runs factory, rebuilds widget
```

So a *partial* update such as `updateHint({{text, "Indexing… 42%"}})` will:
1. reset the icon back to `spec.icon` (discarding an earlier `updateHint({icon,...})`),
2. reset actions back to `spec.actions`,
3. re-invoke the spec's left/right factories (discarding any runtime widget state), and
4. trigger Bug 1 (spacer accumulation) via the right rebuild.

Live updates are not composable — later partial updates silently clobber earlier ones. If the intended contract is "partial patch", the `else if (!content.isEmpty()) / else if (spec...)` branches should be dropped (only apply keys actually present in `content`). If the intended contract is "full refresh with overrides", that should be documented and `handleViewHintUpdate` callers must always pass a complete map. Given the granular `ViewHintUpdateKey` set, partial-patch is the more natural expectation.

---

### 🟡 Should fix 3 — Stale left widget when a factory returns `nullptr` (`viewhintmessage.cpp`)

Both `setCustomWidgetFactory(_, Left)` and `refresh()` guard the left assignment with `if (w)`:

```cpp
if (factory) {
    auto *w = factory(d->message.data());
    if (w)                                       // <- if factory returns null, do nothing
        d->message->setCustomWidget(w, Side::Left);
} else {
    d->message->setCustomWidget(nullptr, Side::Left);
}
```

If a stored left factory is callable but conditionally returns `nullptr` during a live update/refresh, the previously-created left widget is left in place — the display becomes stale instead of reflecting the factory's current result. Drop the `if (w)` and always call `setCustomWidget(w, Left)` (passing `nullptr` clears the old widget, exactly as the null-factory branch does). Same change is needed in `refresh()`.

---

### 🟡 Should fix 4 — Translation context changed; no `.ts`/`.po` updates (`searchhintcontroller.cpp`)

Moving the auth-hint logic from `SearchHelper` to `SearchHintController` changes the `tr()` context for every user-visible string ("Authorize to enable %1", "Enable", "View", "Index is being built…", "File Manager", "You can manually enable %1…", and the mode names). Existing translations live under the `SearchHelper` context, and this PR touches no translation files, so those strings will render in English until re-translated under `SearchHintController`.

Please either preserve the original context with `QCoreApplication::translate("SearchHelper", ...)` (keeps current translations valid), or include the corresponding `.ts`/`.po` updates in this PR.

---

### 🟢 Nits / suggestions

5. **Duplication between `tryShowViewHint` and `handleViewHintShow`.** The hint-creation logic is near-duplicated, and the two empty-content paths diverge (`tryShowViewHint` calls `applyContentToHint` then re-sets icon/text/actions — the icon set is redundant). Also, the only caller of `tryShowViewHint` is `setCurrentView(url)` with no `content`, so the non-empty `content` branch there is currently dead. Consider unifying on one helper.

6. **`SearchHintController::disabledSearchModes()` regressed from `static` to an instance method.** It uses no instance state yet is reached via `instance()->disabledSearchModes()`. Make it `static` again (matching the original `SearchHelper` version) for consistency with the other statics.

7. **Unused infrastructure in this PR.** `SearchHintController::showHint/closeHint/updateHint` and the `ViewHintCustomWidgetFactory` metatype declared in `search.cpp` are not exercised here — the search plugin registers neither a left nor a right factory and never calls `updateHint`. The header comment even advertises "搜索插件设置左侧为 DSpinner 并清除图标", but that wiring isn't present. Either land the DSpinner usage in this PR or defer these helpers (and the metatype) to the PR that actually consumes them, to avoid carrying dead code.

8. **Tests are smoke-only.** The new cases use `EXPECT_NO_FATAL_FAILURE` exclusively, so they assert no behavior — not that the icon actually changed, that the widget landed on the correct side, or that the layout item count is stable. Consequently Bug 1 (spacer leak) is not caught. Consider adding assertions such as checking `msg`'s underlying layout count after repeated rebuilds, and `customWidget(Side::Left/Right)` identity after live updates.

9. **`ViewHintMessage::setCustomWidget(QWidget*, Side)` bypasses the stored factory.** A directly-set widget is silently overwritten by a subsequent `refresh()` (which uses the factory). Either remove this overload or document that it does not update the factory and is not refresh-safe.

10. **Whitespace:** a stray leading space was introduced before two `/*!` doc blocks (`viewhintmessage.h`, `viewhintwidget.h`). Trivial.

---

### Summary

Overall the abstraction is reasonable and well-tested for crashes, but I'd ask for **Bug 1 (spacer leak)** and **Bug 2 (update resets)** to be fixed before merging, plus a decision on **#4 (translations)**. Happy to take another look after updates.

Johnson-zs (approved):

